PR #110, opened 2021-08-24
Add section about setting up binary as a service

anikitinDSR:

### DIFF
--- a/docs/setting-up-a-new-node.md
+++ b/docs/setting-up-a-new-node.md
@@ -146,13 +146,15 @@ You can get the binary in several ways:
 ### Setting up `cheqd-noded` binary as a service
 
 It's highly recommended running node under the service, for example `systemd` system can be used for it. 
-Our debian package uses [postinst](https://github.com/cheqd/cheqd-node/blob/main/build_tools/postinst) script for setting up our binary as a service and it was designed for using outside deb package too. 
-For setting up your binary as a service on ubuntu just run (it requires root privileges):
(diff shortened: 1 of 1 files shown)


docs/setting-up-a-new-node.md
```diff
@@ -143,6 +143,17 @@ You can get the binary in several ways:
...
+### Setting up `cheqd-noded` binary as a service
+
+It's highly recommended running node under the service, for example `systemd` system can be used for it. 
+Our debian package uses [postinst](https://github.com/cheqd/cheqd-node/blob/main/build_tools/postinst) script for setting up our binary as a service and it was designed for using outside deb package too. 
+For setting up your binary as a service on ubuntu just run (it requires root privileges):
+```
+# bash postinst <path/to/cheqd-noded/binary> 
+```
+and this script will add service file and prepare all needed directories for `configs/keys` and `data`.
+Please make sure, it assumes to create and use `cheqd` user for service and all the manipulations with `cheqd-noded` should be run under the `cheqd` user in the future.
```
ankurdotb, line 158: I'm not sure what this comment means...does it mean that postint  **will** create a user called `cheqd`, or should the user previously have created a user called `cheqd`?

Either way, are there any parameters that need to be passed to `postinst` so that it configures directories or to expect/use the `cheqd` user correctly?

ankurdotb: @anikitinDSR Hey could you please take a look at the above and resolve? I also made a commit with some edits, so you might have to sync with that commit first.

anikitinDSR: Agreed, it's not the correct phrase. I meant that this script will create user `cheqd`  cause it's needed for systemd to start with separate from another users environment.

PR #117, opened 2021-09-01
ADR: Fees processing

askolesov:


architecture/adr-list/adr-007-estimating-transaction-fees.md
```diff
@@ -0,0 +1,133 @@
...
+
+`gas-prices` is the price of gas suggested by transaction sender. It is set by the transaction sender and can be set to any value.
+
+Validators can define a parameter called `min-gas-prices`, which acts as a filter on the transactions in each validators' mempool. Transactions with lower `gas-prices` are not considered as candidates for inclusion into the next blocks by the validator. This value is configured by the node/validator operator in the `app.toml` config file.
+
+If `gas-prices` specified in a transaction is lower than the `min-gas-prices` of _all_ validators, no validators consider that transaction in their mempool. The transaction eventually times out and is not processed.
+
+If `gas_prices` specified in a transaction is lower than `min-gas-prices` of _some_ validators, the transaction has a chance at being committed within the transaction timeout window. Whether this actually happens depends on the sum of voting powers of validators that are ready to commit it.
+
+There is currently no way to request `min-gas-prices`.
```
ankurdotb, line 53: The original text said "there is no way to request gas-prices". Since gas-prices are set by the transaction sender, and dynamically at run-time, I don't think they can be "queried". Am I correct in thinking that what you actually meant to say is there's no way to query min-gas-prices, which is a static (as long as it's in the config file) per validator?

architecture/adr-list/adr-007-estimating-transaction-fees.md
```diff
@@ -0,0 +1,133 @@
...
+**Option 1: Client-side calculation**
+
+1. A gas estimation request could be built into the client libraries \(such as [VDR Tools](https://gitlab.com/evernym/verity/vdr-tools)\) used in applications using cheqd network
+2. The client library would calculate an adjustment coefficient client-side.
+
+**Option 2: Use estimated values**
+
+1. Estimated gas for most common transactions could be pre-calculated based on best-guess.
+2. This method can be used as a workaround in the absence of a formal gas estimation request in Cosmos.
+3. This scenario is more likely to fail over time as gas prices can be dynamic on a live network with lots of transactions.
```
ankurdotb, line 74: IMO, this option is a no-go. We can estimate this values, but the moment mainnet is live (or even testnet with any significant volume), it's a free-for-all. Any validator could set any gas wanted or min gas price and at that point all estimates go out of the window. Would you agree?

architecture/adr-list/adr-007-estimating-transaction-fees.md
```diff
@@ -0,0 +1,133 @@
...
+`gas` above must be at least equal to or more than `gas-wanted`. 
+
+### Options considered for estimating transaction fees
+
+#### Gas estimation
+
+**Option 1: Client-side calculation**
+
+1. A gas estimation request could be built into the client libraries \(such as [VDR Tools](https://gitlab.com/evernym/verity/vdr-tools)\) used in applications using cheqd network
+2. The client library would calculate an adjustment coefficient client-side.
```
ankurdotb, line 68: Yes, let's use a weighted moving average (more on this later). Doing this client-side might mean needing a light client, or by querying a backend service.

architecture/adr-list/adr-007-estimating-transaction-fees.md
```diff
@@ -0,0 +1,133 @@
...
+1. Estimated gas for most common transactions could be pre-calculated based on best-guess.
+2. This method can be used as a workaround in the absence of a formal gas estimation request in Cosmos.
+3. This scenario is more likely to fail over time as gas prices can be dynamic on a live network with lots of transactions.
+
+#### Fee price estimations
+
+**Option 1: Use fixed values**
+
+1. Set a fixed recommended gas price on the cheqd network
+2. Embed the fixed value into applications
```
ankurdotb, line 81: For similar reasons as above, we cannot count on this.

Not all transactions will be from SSI wallet apps using VDR Tools. In fact, the bulk of the transaction activity will likely be just pure crypto trading. We cannot guarantee which wallet app(s) any token holder would be using, and therefore we have no ability to embed a fixed value into applications.

This method ONLY works if we're considering Connect.me/Verity as the only applications sending/receiving requests. That is not the scenario unfortunately, even for the v1 of mainnet.

architecture/adr-list/adr-007-estimating-transaction-fees.md
```diff
@@ -0,0 +1,133 @@
...
+
+**Option 2: Dynamically calculate values**
+
+1. Gas prices can be dynamically determined client side based on recent transactions.
+2. This could be implemented as a client-side calculation, provided as a service \(e.g., via an oracle\), or a blend of both approaches.
```
ankurdotb, line 86: Same as above, potentially a weighted moving average?

ankurdotb (changes requested):
Back to you @askolesov, a couple of notes:

1. Seems like we haven't actually agreed on a method, so I've set status to DRAFT and moved the Decision section into an options considered scenario instead.
2. Please see review comments through as well. Some are easily addressed.
3. Primarily, I'm interested in the dynamic vs fixed values idea. IMO, there's no way in a viable mainnet we can have fixed values, as we cannot guarantee all uses will be using apps made by us.
4. On how we do dynamic calculations: I would suggest either a weighted moving average or exponential moving average. There are likely implementations available for this already rather than having to re-implement from scratch. I assume you know these terms but just in case: https://en.wikipedia.org/wiki/Moving_average
5. While we can't rely on fixed values, we don't need to solve fee estimation for all kinds of client apps: for now. If a token holder (purely crypto, no SSI credentials) wants to use their own client app to hold and transact tokens, that's fine: it's upto their preferred client application to estimate gas/gas prices.
6. Therefore, we only need to solve fee estimation for Connect.me. Doing this client-side may require implementing a Tendermint light client, which is less-than-ideal because of the effort involved. What I would suggest instead is that the gas/gas prices are computed as an Evernym backend service for Connect.me, regularly pre-compute gas/gas price values every X time window, and the client app can either fetch/push-receive those figures if/when it tries to initiate a send transaction.
7. If we don't want to lock people into an Evernym backend with their client apps, fair. For that I would say, this backend service to pre-compute gas/gas prices could also be open sourced. It could also, for instance, be a Lambda function. We'll have a firehose of tx data once the network gets going: the Lamba could be initiated every X minutes/hours, compute the figure, spit out a file. We could even make this smart enough to dynamically vary the time window when this function runs, based on how stable/unstable the values are. OR it could be a live compute for every single request: I really don't mind, this is really more down to the Verity/Connect.me teams. Point is, we can release this so that anyone else implementing similar SSI wallet apps can also use it.

On that note...how do other networks/crypto wallet apps for Cosmos networks do this? Do they compute client-side or server-side? Does anybody use the pre-computation method I described above?

architecture/adr-list/adr-007-estimating-transaction-fees.md
askolesov: Correct, my bad wording here

architecture/adr-list/adr-007-estimating-transaction-fees.md
askolesov: Gas is pretty constant for simple transactions (aka computation complexity) and can't be changed. It can fluctuate slightly depending on the database state. Gas price isn't constant.

> Any validator could set any gas wanted

No, gas wanted is set by the sender. Gas used is computed during execution. No way for a validator operator to influence on it.

In general, agreed, I don't like this option either although it should work. Just wanted to include it as an option in the scope of branstorm.

architecture/adr-list/adr-007-estimating-transaction-fees.md
askolesov: It can be resource-intensive for mobiles. Our latest idea is to do it on-chain which is much better in all aspects. For example, it will support state proofs.

architecture/adr-list/adr-007-estimating-transaction-fees.md
askolesov: Agreed, again, it was a part of the brainstorm.

architecture/adr-list/adr-007-estimating-transaction-fees.md
askolesov: Yep, moving average is a great idea

askolesov:
Thanks for the review!

> 1. Seems like we haven't actually agreed on a method, so I've set status to DRAFT and moved the Decision section into an options considered scenario instead.

Agreed

> 2. Please see review comments through as well. Some are easily addressed.

Done

> 3. Primarily, I'm interested in the dynamic vs fixed values idea. IMO, there's no way in a viable mainnet we can have fixed values, as we cannot guarantee all uses will be using apps made by us.

Agreed

> 4. On how we do dynamic calculations: I would suggest either a weighted moving average or exponential moving average. There are likely implementations available for this already rather than having to re-implement from scratch. I assume you know these terms but just in case: https://en.wikipedia.org/wiki/Moving_average

Great idea indeed. In the case of on-chain computation, we have to reimplement it anyway taking into account a lot of blockchain-related nuances.

> 5. While we can't rely on fixed values, we don't need to solve fee estimation for all kinds of client apps: for now. If a token holder (purely crypto, no SSI credentials) wants to use their own client app to hold and transact tokens, that's fine: it's upto their preferred client application to estimate gas/gas prices.

At least 2 apps will use VDR library: connect.me and verity. We can implement estimation inside it.

> 6. Therefore, we only need to solve fee estimation for Connect.me. Doing this client-side may require implementing a Tendermint light client, which is less-than-ideal because of the effort involved. What I would suggest instead is that the gas/gas prices are computed as an Evernym backend service for Connect.me, regularly pre-compute gas/gas price values every X time window, and the client app can either fetch/push-receive those figures if/when it tries to initiate a send transaction.

As far as I know, there is no back-end service for Connect me except verity. On-chain computation should solve this.

> 7. If we don't want to lock people into an Evernym backend with their client apps, fair. For that I would say, this backend service to pre-compute gas/gas prices could also be open sourced. It could also, for instance, be a Lambda function. We'll have a firehose of tx data once the network gets going: the Lamba could be initiated every X minutes/hours, compute the figure, spit out a file. We could even make this smart enough to dynamically vary the time window when this function runs, based on how stable/unstable the values are. OR it could be a live compute for every single request: I really don't mind, this is really more down to the Verity/Connect.me teams. Point is, we can release this so that anyone else implementing similar SSI wallet apps can also use it.

Interesting option.

> On that note...how do other networks/crypto wallet apps for Cosmos networks do this? Do they compute client-side or server-side? Does anybody use the pre-computation method I described above?

I didn't see any of the implementations. Cosmos CLI delegates gas and fees estimation up to user.

askolesov:
I've committed the latest thoughts. More thoughts are coming based on our discussions.

architecture/adr-list/adr-008-estimating-transaction-fees.md
```diff
@@ -0,0 +1,253 @@
...
+1. Gas prices can be dynamically determined based on recent transactions.
+2. Possible ways of estimation:
+   1. Moving average
+   2. Weighed moving average
+   3. Exponential moving average
+3. Possible ways of implementation:
+   1. Client side
+   2. Provided as a service
+   3. On chain (the most trusted and robust way)
+3. To prevent constant growth of the mean value it's proposed to use multiplier `Y < 1` to get star fee price.
```
ankurdotb, line 95: Unsure what is meant by "star fee price"?

architecture/adr-list/adr-008-estimating-transaction-fees.md
```diff
@@ -0,0 +1,253 @@
...
+4. If the dynamically calculated fees are insufficient to put a transaction though, use exponential growth to calculate higher values and retry.
+
+### Proposed transaction sending flow in VDR Tools
+
+1. Build and sign a transaction
+2. Send gas estimation request
+   1. Multiply response to the precalculated coefficient `X > 1` to increase probability of the transaction being committed
```
ankurdotb, line 102: The cascading impact of this though is this will also very quickly cause the gas estimates to move upwards. I'd suggest trying this only if a transaction fails.

architecture/adr-list/adr-008-estimating-transaction-fees.md
```diff
@@ -0,0 +1,253 @@
...
+3. Set initial gas price
+   1. Request moving average from the network
+   2. Multiply response by `Y < 1` to prevent constant growth of the mean value
```
ankurdotb, line 105: Yes, this is a good idea. Although how do you want to set this multiplier/factor...random? Or other metric?

architecture/adr-list/adr-008-estimating-transaction-fees.md
```diff
@@ -0,0 +1,253 @@
...
+   2. Exponentially increase the `gas-prices`, in case transaction time-out \(as this indicates `min-gas-prices` was not met\)
+
+### Proposed API in VDR Tools
+
+#### Build simulate query
+
+Used to build gas estimation request. 
+
+```
+indy_cheqd_ledger_tx_build_query_simulate(tx: byte[]) -> String
```
ankurdotb, line 117: If the API changes, these function names are also likely to change

architecture/adr-list/adr-008-estimating-transaction-fees.md
```diff
@@ -0,0 +1,253 @@
...
+```
+
+Parameters:
+* `tx` - transaction bytes from `build_tx` method
+
+Result:
+* request body
+
+`indy_cheqd_pool_abci_query` should be used to send the request and receive a response.
+
```
ankurdotb, line 127: Same here, if the APIs change these names will likely change

architecture/adr-list/adr-008-estimating-transaction-fees.md
```diff
@@ -0,0 +1,253 @@
...
+Return string is json encoded structure:
+
+```
+PricesResponse {
+   LastAverage uint64
+   SubsetSize uint64
+}
+```
+
+`LastAverage` - average gas price for the recent `N` committed transactions where `N` is `SubsetSize`.
```
ankurdotb, line 204: A very large or very small value will move this figure too much potentially out of range

ankurdotb (commented):
I'm not a 100% sure from the text of this ADR whether the proposal is to calculate these values client-side only, or whether there are some calculations done on ledger.

Broadly I like the ideas proposed here, BUT I see one big issue with doing computation on ledger: one or more set of validators could band together to deliberately jack up gas-wanted or min-gas-prices. They already know how the fee estimation is carried out, what window of transactions to target, and what kind of average/growth mechanisms are used. 

Right now if they did that, they might price themselves out of having enough transactions in their mempool. However, if the calculation is done on chain, then they know with a high degree of confidence that a lot of client apps (mobile or backend) will follow the lead of what figures are being returned from on-chain. They'll know that anyone using VDR Tools SDK will follow their lead from whatever they start feeding on-chain.

For this reason, I think the calculation should be carried out on the web backend, i.e., Verity. If the estimates are off, then at least only the client apps/web app backends that use the same estimation strategy are affected...and they can always use exponential growth or other strategies to catchup.

ankurdotb:
> As far as I know, there is no back-end service for Connect me except verity. On-chain computation should solve this.

What I meant is Verity itself is likely composed of microservices, not a monolith. So there could be a fee estimation microservice.

> > 1. If we don't want to lock people into an Evernym backend with their client apps, fair. For that I would say, this backend service to pre-compute gas/gas prices could also be open sourced. It could also, for instance, be a Lambda function. We'll have a firehose of tx data once the network gets going: the Lamba could be initiated every X minutes/hours, compute the figure, spit out a file. We could even make this smart enough to dynamically vary the time window when this function runs, based on how stable/unstable the values are. OR it could be a live compute for every single request: I really don't mind, this is really more down to the Verity/Connect.me teams. Point is, we can release this so that anyone else implementing similar SSI wallet apps can also use it.
> 
> Interesting option.

An example of how a fee estimation microservice could be implemented, without really touching the rest of Verity.

Also note that fee estimation along with identity software is only relevant for Evernym Verity/VDR Tools. If the fee estimation service was written like a microservice above, it's also usable by any token-only client apps, without needing full Verity suite.

> > On that note...how do other networks/crypto wallet apps for Cosmos networks do this? Do they compute client-side or server-side? Does anybody use the pre-computation method I described above?
> 
> I didn't see any of the implementations. Cosmos CLI delegates gas and fees estimation up to user.

Lastly, I want to check if we're over-engineering this. If other Cosmos networks work fine without this, presumably they do a few naive calculations + exponential growth and call it a day. Why not do similar?

architecture/adr-list/adr-008-estimating-transaction-fees.md
askolesov: If we implement a retry strategy it will accept `start gas price`, `max gas price`, and `number of retries`. Those are other questions - how will it be shown on UI and what API are we going to provide here.

architecture/adr-list/adr-008-estimating-transaction-fees.md
askolesov: At this point the proposal is to take either minimum or let's say 5th percentile and multiply it by 0.9. Numbers are taken from the top of my head and are subjects of research or experiment.

architecture/adr-list/adr-008-estimating-transaction-fees.md
askolesov: Agree

architecture/adr-list/adr-008-estimating-transaction-fees.md
askolesov: The strategy should be chosen by a particular app/its back-end.

askolesov:
> Broadly I like the ideas proposed here, BUT I see one big issue with doing computation on ledger: one or more set of validators could band together to deliberately jack up gas-wanted or min-gas-prices. They already know how the fee estimation is carried out, what window of transactions to target, and what kind of average/growth mechanisms are used.

True, that's our concern too. By the way, `gas-wanted` or `gas-used` are not subjects of attack. They are deterministic. The only possible attack that we thought about is lying with the gas estimate.

> For this reason, I think the calculation should be carried out on the web backend, i.e., Verity. If the estimates are off, then at least only the client apps/web app backends that use the same estimation strategy are affected...and they can always use exponential growth or other strategies to catchup.

Totally agree.

> > > On that note...how do other networks/crypto wallet apps for Cosmos networks do this? Do they compute client-side or server-side? Does anybody use the pre-computation method I described above?
> > 
> > I didn't see any of the implementations. Cosmos CLI delegates gas and fees estimation up to user.
> 
> Lastly, I want to check if we're over-engineering this. If other Cosmos networks work fine without this, presumably they do a few naive calculations + exponential growth and call it a day. Why not do similar?
> 
> I'd also say this is a perfect example of a question we can ask All In Bits team. Let's summarise our approaches/thoughts and ask today?

Would be perfect.

So the list of questions we have:
* Where to implement gas price estimation?
  * Best option right now is a separate backend. Could also be inside libindy.
Questions that are more for Evernym:
* How fee will be shown on UI in connect.me?

ankurdotb:
TBD: do something similar to Keplr https://app.clickup.com/t/6600954/DEV-434

Toktar:
Updated PR.
@ankurdotb Could you please review it.

ankurdotb:
I'm gonna close this a WONTDO, since most other networks just use pre-defined values that work with a variety of Cosmos wallets.

PR #133, opened 2021-09-10
Documentation about API for keys and pool methods.

anikitinDSR:


askolesov:
@ankurdotb Looks like there will be several ADRs that are cheqd related but about VDR Tools. What is the right place for them?

ankurdotb:
> @ankurdotb Looks like there will be several ADRs that are cheqd related but about VDR Tools. What is the right place for them?

I don't see this document as an ADR, it's more a developer documentation for how to use the software. Good question on whether it should be here though: if it's about how to use VDR Tools, then I'd expect it to be im VDR Docs but we can link to those docs from ours. Is there doc site similar to docs.cheqd.io for VDR Tools?

docs/pool_connection.md
```diff
@@ -0,0 +1,246 @@
...
+This page describes API about how to work with keys and how to connect to the pool. 
+All the libraries methods/calls are defined on the [`cheqd-sdk`](https://github.com/cheqd/cheqd-sdk)    (forked from [Evernym VDR Tools](https://gitlab.com/evernym/verity/vdr-tools) ).
+
+## Keys
+### List of methods
+* [`indy_cheqd_keys_add_random`](#indy_cheqd_keys_add_random)
+* [`indy_cheqd_keys_add_from_mnemonic`](#indy_cheqd_keys_add_from_mnemonic)
+* [`indy_cheqd_keys_get_info`](#indy_cheqd_keys_get_info)
+* [`indy_cheqd_keys_get_list_keys`](#indy_cheqd_keys_get_list_keys)
+* [`indy_cheqd_keys_sign`](#indy_cheqd_keys_sign)
```
ankurdotb, line 14: Do these names/methods now change under the unified API?

docs/pool_connection.md
```diff
@@ -0,0 +1,246 @@
...
+```
+
+#### **indy_cheqd_keys_add_from_mnemonic**
+This method realised logic for recovering keys from `mnemonic` string. Mnemonic string - it's a human-readable combination of words, in general can include 12 or 24 words.
+
+Input parameters:
+
+* `wallet_handle` - integer, which is connected to previously created and opened wallet.
+* `alias` - human-readable representation of alias to user,
+* `mnemonic` - string of 12 or 24 words.
```
ankurdotb, line 40: Can we show examples of what a request looks like? Output is fine.

docs/pool_connection.md
```diff
@@ -0,0 +1,246 @@
...
+}), 
+Object({
+	"alias": String("test_pool_2"),
+	"chain_id": String("chain_id"),
+	"rpc_address": String("rpc_address")
+})]
+```
+#### **indy_cheqd_pool_broadcast_tx_commit**
+This method allows to send a txn to all the nodes.
+
```
ankurdotb, line 155: Do we really need this? It'll be impossible for a node to know where ALL nodes are, at best it will be able to know the peers connected to it. Cosmos handles this by default, so why do we need a separate call?

docs/pool_connection.md
```diff
@@ -0,0 +1,246 @@
...
+* [`indy_cheqd_pool_abci_query`](#indy_cheqd_pool_abci_query)
+* [`indy_cheqd_pool_abci_info`](#indy_cheqd_pool_abci_info)
+
+#### **indy_cheqd_pool_add**
+This method is needed for adding information about pool which will be used to connect to.
+
+Input parameters:
+* `alias` - is a human-readable string,
+* `rpc_address` - address for connecting to the node, like `http://1.2.3.4:26657`, port `26657` is default value.
+* `chain_id` - identifier of the network.
```
ankurdotb, line 115: What is the purpose of this, if a node has already been added using cheqd Cosmos CLI?

docs/pool_connection.md
```diff
@@ -0,0 +1,246 @@
...
+```
+{
+    "alias":"test_pool",
+    "rpc_address":"rpc_address",
+    "chain_id":"chain_id"
+}
+```
+#### **indy_cheqd_pool_get_config**
+This method is needed for getting config information about connecting to the pool by alias.
+
```
ankurdotb, line 126: Same as above: this is in cheqd Cosmos CLI, so why replicate it here?

docs/pool_connection.md
```diff
@@ -0,0 +1,246 @@
...
+Expected result is structure PoolConfig, like:
+```
+{
+    "alias":"test_pool",
+    "rpc_address":"rpc_address",
+    "chain_id":"chain_id"
+}
+```
+#### **indy_cheqd_pool_get_all_config**
+The same as [indy_cheqd_pool_get_config](#indy_cheqd_pool_get_config) but returns the list of structures.
```
ankurdotb, line 139: Same as above

docs/pool_connection.md
ankurdotb: Also with the way that Cosmos works, if the transaction is submitted to multiple validator nodes:

1. Will it not just appear like a duplicate transaction was attempted?
2. Is it not going to cost a LOT of gas/fees to pay to multiple nodes>

ankurdotb (changes requested):
We should discuss this...it seems there are things being implemented here that are duplicates of what cheqd Cosmos CLI is meant to achieve. To me, this would go against what we decided in the ADR for CLI tools: https://docs.cheqd.io/node/architecture/adr-list/adr-003-cli-tools

More broadly though, VDR Tools SDK should assume the network is already up and running, and no configuration tasks are expected to be carried out from VDR Tools. Broadcast as a concept also doesn't make sense to me.

ankurdotb (approved):

PR #140, opened 2021-09-22
Identity implementation

Toktar:
TODO:
**Return cosmos-sdk v0.42.5 !!!**


askolesov:
Ok to merge with follow-up tickets created from the list of concerns above.

PR #142, opened 2021-09-23
Cosmos v0.44

askolesov:


docker/cheqd_node/Dockerfile
```diff
@@ -31,18 +31,21 @@ RUN starport chain build
...
+RUN groupadd --system --gid 1000 cheqd && \
+    useradd --system --create-home --home-dir /cheqd --shell /bin/bash --gid cheqd --uid 1000 cheqd
+RUN chown -R cheqd /cheqd
 
-WORKDIR /home/cheqd
+WORKDIR /cheqd
 USER cheqd
 
-VOLUME /home/cheqd/.cheqdnode
+VOLUME /cheqd
```
anikitinDSR, line 48: How this change will affect single_image_test building? Previously there was a directory `.cheqdnode` inside the WORKDIR, but for now it does not.

docker/single_image_testnet/Dockerfile
```diff
@@ -17,10 +17,9 @@ EXPOSE 26657 26659
 RUN ./gen_node_configs.sh
 
 # Update persistent peers list for at least one node
-RUN NODE_0_ID=$(cheqd-noded tendermint show-node-id --home node_configs/node0); \
-    NODE_1_ID=$(cheqd-noded tendermint show-node-id --home node_configs/node1); \
-    PERSISTENT_PEERS="$NODE_0_ID@127.0.0.1:26656,$NODE_1_ID@127.0.0.1:26658"; \
-    sed -i $extension "s/persistent_peers = \"\"/persistent_peers = \"$PERSISTENT_PEERS\"/g" node_configs/node0/config/config.toml
+RUN NODE_0_ID=$(cheqd-noded tendermint show-node-id); \
+    PERSISTENT_PEERS="$NODE_0_ID@127.0.0.1:26656"; \
+    sed -i $extension "s/persistent_peers = \"\"/persistent_peers = \"$PERSISTENT_PEERS\"/g" .cheqdnode/config/config.toml
```
anikitinDSR, line 22: And this step expects that there is a directory in the WORKDIR, which is inherited from `cheqd-node` image

test/networks/docker_compose/docker-compose.yml
```diff
@@ -0,0 +1,56 @@
+version: '3.2'
+
+services:
+  node0:
+    image: cheqd-node
+    ports:
+      - "26656:26656" # p2p
+      - "26657:26657" # rpc
+    volumes:
+      - ./node_configs/node0:/cheqd:Z
```
anikitinDSR, line 10: The same question. It mounts directory `node0` into `/cheqd` but does node0 contains '.cheqdnode' inside?

anikitinDSR (approved):

app/app.go
```diff
@@ -84,13 +96,11 @@ import (
...
 	upgradetypes "github.com/cosmos/cosmos-sdk/x/upgrade/types"
 	tmjson "github.com/tendermint/tendermint/libs/json"
-	tmproto "github.com/tendermint/tendermint/proto/tendermint/types"
 	// this line is used by starport scaffolding # stargate/app/moduleImport
 )
 
-const Name = "cheqd-node"
-const Version = "1.0"
-const AppVersion = 1
+const Name = "cheqdnode"
```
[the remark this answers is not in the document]
askolesov: Good catch! My mistake during conflict resolution.

docker/cheqd_node/Dockerfile
askolesov: It's just the home directory change. I fixed the testnet image as well.

docker/single_image_testnet/Dockerfile
askolesov: This change is because previously we had several nodes inside the mage and their homes were `node_configs/node0`, `node_configs/node1`, and so on. Now there is only one node with the home in `.cheqdnode`.

test/networks/docker_compose/docker-compose.yml
askolesov: Yep, I changed Config generation script as well.

PR #143, opened 2021-09-23
Add DIDDoc to Identity ADR

Toktar:


architecture/adr-list/adr_002_identity_transactions.md
```diff
@@ -44,103 +44,191 @@ _**Note**: Hyperledger Indy also contains other transaction types beyond the one
...
 
 ### `DID` transactions
 
 [Decentralized Identifiers \(DIDs\) are a W3C specification](https://www.w3.org/TR/did-core/) for identifiers that enable verifiable, decentralized digital identity.
+DIDDoc format conforms to [DIDDoc spec]().
+The request can be used for creation of new DIDDoc, setting, and rotation of verification key.
 
-The request can be used for creation of new DIDs, setting, and rotation of verification key.
+#### DIDDoc
+1. **`id` (base58-encoded string):**
```
asmarty, line 97: I know this has been discussed in meetings, BUT my point is that if we implement the DID Doc and store them in the ledger, this document must conform to the DID Doc specification. I am not sure if it is correct to convert this structure to DID Doc in VDR tool. What happens if someone implements another client and suspects the ledger will return the DID Doc?

architecture/adr-list/adr_002_identity_transactions.md
```diff
@@ -44,103 +44,191 @@ _**Note**: Hyperledger Indy also contains other transaction types beyond the one
...
+2. **`controller` (list of base58-encoded string):**
+
+   A set of identifier strings.
+
+3. **`type` (string).**
+4. **`publicKeyJwk` (map[string,string], optional):**
+
+   A map representing a JSON Web Key that conforms to [RFC7517](https://tools.ietf.org/html/rfc7517). See definition of publicKeyJwk for additional constraints.
+
+5. **`publicKeyMultibase` (base58-encoded string, optional):**
```
asmarty, line 178: I would prefer to see publicKeyBase58 as well.

asmarty: please ignore this. as I understand https://www.w3.org/TR/did-core/#revision-history `publicKeyBase58` replaced by  `publicKeyMultibase`.

architecture/adr-list/adr_002_identity_transactions.md
```diff
@@ -44,103 +44,191 @@ _**Note**: Hyperledger Indy also contains other transaction types beyond the one
...
-```text
-Did
-{
-    "alias": "Alice DID",
-    "dest": "GEzcdDLhCpGCYRHW82kjHd",
-    "verkey": "~HmUWn928bnFT6Ephf65YXv"
-}
+2. **`controller` (list of base58-encoded string):**
+
+   A set of identifier strings.
```
askolesov, line 171:
```suggestion
   An identifier strings.
```

architecture/adr-list/adr_002_identity_transactions.md
```diff
@@ -44,103 +44,191 @@ _**Note**: Hyperledger Indy also contains other transaction types beyond the one
...
+        "service": [{
+        "id":"N22KY2Dyvmuu2PyyqSFKue#linked-domain",
+        "type": "LinkedDomains",
+        "serviceEndpoint": "https://bar.example.com"
+        }]
+    ````
+
+#### **Update DID**
+
+If there is no DID transaction with the specified DID \(`dest`\), it is considered as a creation request for a new DID.
```
askolesov, line 223: `dest` looks like an outdated field.

architecture/adr-list/adr_002_identity_transactions.md
```diff
@@ -44,103 +44,191 @@ _**Note**: Hyperledger Indy also contains other transaction types beyond the one
...
-**Note:** ATTRIB **can** be updated
+    ```
+        "service": [{
+        "id":"N22KY2Dyvmuu2PyyqSFKue#linked-domain",
+        "type": "LinkedDomains",
+        "serviceEndpoint": "https://bar.example.com"
+        }]
+    ````
+
+#### **Update DID**
```
askolesov, line 221: It's not the cosmos way. I highly recommend split it into create/update transactions.

askolesov (commented):
Questions:
- Do we allow external refs?
- Do we validate DIDs?
- How are we going to handle reply protection?
  - Is it allowed to add arbitrary files in diddoc?
  - We need two: chain-id and seq-no.
  - If we stored full dids on-chain, there was no need for chain-id.

architecture/adr-list/adr_002_identity_transactions.md
```diff
@@ -44,103 +44,191 @@ _**Note**: Hyperledger Indy also contains other transaction types beyond the one
...
 ## List of transactions and details
 
 ### `DID` transactions
 
 [Decentralized Identifiers \(DIDs\) are a W3C specification](https://www.w3.org/TR/did-core/) for identifiers that enable verifiable, decentralized digital identity.
+DIDDoc format conforms to [DIDDoc spec]().
+The request can be used for creation of new DIDDoc, setting, and rotation of verification key.
 
-The request can be used for creation of new DIDs, setting, and rotation of verification key.
+#### DIDDoc
```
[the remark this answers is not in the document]
asmarty: Join @askolesov

architecture/adr-list/adr_002_identity_transactions.md
Toktar: It should be a single string. Sorry, my mistake, it needs to be fixed.

architecture/adr-list/adr_002_identity_transactions.md
Toktar: My concerns:
- How are we going to validate this prefix? With hard-coded values?
- What are we going to do after changing this prefix? Is it possible?

asmarty: I guess if prefix has changed it means did method has changed as well, so it's okey for me hard-coded them.

ankurdotb: We should store the full DID, as in, `did:cheqd:network:identifier` in the body of this request. What the DID method is called and structure is up to us to define. If this is NOT on ledger, then the interpretation of which method is being used, what network, and what structure to expect is happening entirely on a backend/client application.

Not storing the full DID, as far as I understand, violates the DID Core specificiation.

architecture/adr-list/adr_002_identity_transactions.md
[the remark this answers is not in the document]
ankurdotb: Neither of these two are required properties, but fair point on including them just for extensibility in the future.

architecture/adr-list/adr_002_identity_transactions.md
ankurdotb: Yes I would agree. This is from the old request format.

architecture/adr-list/adr_002_identity_transactions.md
ankurdotb: See implementation here: https://github.com/allinbits/cosmos-cash/blob/main/docs/Explanation/ADR/adr-004-did.md#update

architecture/adr-list/adr-002-identity-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+
+1. **`id`**: Target DID as base58-encoded string for 16 or 32 byte DID value. In the ledger, we store only an identifier without specifying the method and namespace.
+2. **`controller`** (optional): A list of base58-encoded identifier strings.
+3. **`verificationMethod`** (optional): A list of Verification Methods
+4. **`authentication`** (optional): A list of Verification Methods or strings with key aliases
+5. **`assertionMethod`** (optional): A list of Verification Methods or strings with key aliases
+6. **`capabilityInvocation`** (optional): A list of Verification Methods or strings with key aliases
+7. **`capabilityDelegation`** (optional): A list of Verification Methods or strings with key aliases
+8. **`service`** (optional): A set of Service Endpoint maps
+9. **`@context`** (optional): A list of strings
```
ankurdotb, line 106: This has a relatively stable definition within DID spec as it typically refers back to W3C DID standards, and can therefore be referred to as such.

architecture/adr-list/adr-002-identity-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+  "publicKeyJwk": {
+    "kty": "OKP", // external (property name)
+    "crv": "Ed25519", // external (property name)
+    "x": "VCpo2LMLhn6iWku8MKvSLg2ZAoC-nlOyPVQaO3FxVeQ" // external (property name)
+}
+```
+
+#### Service
+
+1. **`id`** (string): The value of the id property MUST be a URI conforming to [RFC3986](https://www.rfc-editor.org/rfc/rfc3986). A conforming producer MUST NOT produce multiple service entries with the same ID. A conforming consumer MUST produce an error if it detects multiple service entries with the same ID. It has a follow format: `<DIDDoc-id>#<service-alias>`
```
ankurdotb, line 176: No, a service doesn't need to follow always being a DIDDoc at the start because it's possible to have non-DID services.

architecture/adr-list/adr-002-identity-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+
+If there is no DID transaction with the specified DID \(`dest`\), it is considered as a creation request for a new DID.
+
+If there is a DID transaction with the specified DID \(`dest`\), then this is update of existing DID. In this case, we can specify only the values we would like to override. All unspecified values remain the same. E.g., if a key rotation needs to be performed, the owner of the DID needs to send a DID transaction request with `dest`, `verkey` only. `alias` will stay the same.
+
+**Note**: Fields `dest` and `owner` should have the same value.
+
+#### State format
+
+`id -> {encode(data, creators), tx_hash, tx_timestamp }`
```
ankurdotb, line 200: Already commented in a different comment but this whole section is likely redundant now and needs to be thought of differently. See Cosmos Cash example.

architecture/adr-list/adr-002-identity-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+
+### `SCHEMA`
+
+This transaction is used to create a Schema associated with credentials.
+
```
ankurdotb, line 205: A schema *may* be stored on ledger but it doesn't have to be. For instance, it could be stored as a schema.org URL. Question is...from a wider Evernym perspective, do we want to be flexible here? For instance, on ledger we could allow for this to be either on ledger or somewhere else, and expect the VDR Tools SDK to always read/write from ledger.

architecture/adr-list/adr-002-identity-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+#### `SCHEMA` transaction format
+
+```javascript
+{
+  "id": "N22KY2Dyvmuu2PyyqSFKue",
+  "version": "1.0",
+  "name": "Degree",
+  "attr_names": ["undergrad", "last_name", "first_name", "birth_date", "postgrad", "expiry_date"]
+}
+```
```
ankurdotb, line 225: This should be stored as a DIDDoc, and as defined here it's not stored in a DIDDoc compliant fashion. The format must therefore be changed.

One consequence therefore is much like a DIDDoc therefore, it has the same fields pretty much as other DIDocs (verification method, controllers, etc) and can be updated. Question is: can the older versions be fetched?

architecture/adr-list/adr-002-identity-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+
+#### `SCHEMA` State format
+
+`id -> {encode(data, creators), tx_hash, tx_timestamp }`
+
+
+### `CRED_DEF`
+
+Adds a Credential Definition (in particular, public key), which is created by an Issuer and published for a particular Credential Schema.
+
```
ankurdotb, line 237: Same as schemas above, this is not actually being defined as a DIDDoc here.

ankurdotb (changes requested):
Thanks for the work here @Toktar and for the comprehensive reviews @askolesov @Artemkaaas 

As I see it: 
1. Schemas and Cred Defs need to account for being actually stored as DIDDocs. The current ADR doesn't meet that.
2. I can increasingly see that the [Cosmos Cash implementation](https://github.com/allinbits/cosmos-cash/blob/main/docs/Explanation/ADR/adr-004-did.md) is more compliant with DID specs. So we should consider whether we take their stuff, and then DIDDocs for Schemas and Cred Defs on top.

architecture/adr-list/adr-002-identity-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+        "<public_key>": "<signature>",
+      ],
+    "metadata": {}
+}
+```
+
+- **`data`**: Data requested to be written to the ledger, specific for each request type.
+- **`creators`**: Creators identifiers (DID) list for this entity. There should be a new DIDs or an existing DIDs, for existing entities.
+- **`signatures`**: `data` should be signed by all `creators` private key. This field contains a dict there creator's public key is a key, and the signature is a value.
+- **`metadata`**: Dictionary with additional metadata fields. Empty for now. This fields provides extensibility in the future, e.g., it can contain `protocolVersion` or other relevant metadata associated with a request.
```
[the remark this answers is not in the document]
Toktar: It's not a DIDDoc metadata. It's a write request metadata for every request.

architecture/adr-list/adr-002-identity-transactions.md
```diff
@@ -0,0 +1,274 @@
...
+
+DIDDoc format conforms to [DIDDoc spec](https://www.w3.org/TR/did-core/#representations).
+The request can be used for creation of new DIDDoc, setting, and rotation of verification key.
+
+#### DIDDoc
+
+1. **`id`**: Target DID as base58-encoded string for 16 or 32 byte DID value. In the ledger, we store only an identifier without specifying the method and namespace.
+2. **`controller`** (optional): A list of base58-encoded identifier strings.
+3. **`verificationMethod`** (optional): A list of Verification Methods
+4. **`authentication`** (optional): A list of Verification Methods or strings with key aliases
```
anikitinDSR, line 101: Do we want to use here just list of string as it's implemented in cosmos-cash?
https://github.com/allinbits/cosmos-cash/blob/e8ffcc38d76111a0fda29266f36e364b645531eb/proto/did/did.proto#L44

FYI, @ankurdotb

architecture/adr-list/adr-002-identity-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+
+[Decentralized Identifiers \(DIDs\) are a W3C specification](https://www.w3.org/TR/did-core/) for identifiers that enable verifiable, decentralized digital identity.
+
+DIDDoc format conforms to [DIDDoc spec](https://www.w3.org/TR/did-core/#representations).
+The request can be used for creation of new DIDDoc, setting, and rotation of verification key.
+
+#### DIDDoc
+
+1. **`id`**: Target DID as base58-encoded string for 16 or 32 byte DID value. In the ledger, we store only an identifier without specifying the method and namespace.
+2. **`controller`** (optional): A list of base58-encoded identifier strings.
```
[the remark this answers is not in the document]
Toktar: > "A list of base58-encoded identifier strings" 
allows DID doc to have multiple controllers. Need I add more information to ADR here?

architecture/adr-list/adr-002-identity-transactions.md
Toktar: I extended a description here. But the format of this field is flexible in DIDDoc. A list of any strings provides it.

architecture/adr-list/adr-002-identity-transactions.md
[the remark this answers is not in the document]
Toktar: I didn't add `didDocumentMetadata` and `didResolutionMetadata` because I didn't find cases then they are needed. And  `deactivated` field. I think this field is superfluous, since we don't have such a business requirement. And according to the experience of working on entities that act for a certain amount of time (TAA, revocations), this raises many legal questions for business and non-trivial nuances of implementation.
Please correct me if these structures are required. Will consider the possible options.

architecture/adr-list/adr-002-identity-transactions.md
Toktar: Cosmos Cash describes DIDDoc and doesn't tell about ways of storage. I suggest leaving it here because it's and ADR about transactions. Not only protocol.
But I think we can remove creators from value because `data` already contains this.

architecture/adr-list/adr-002-identity-transactions.md
Toktar: My opinion, that the main feature of Cheqd it's an opportunity of storing everything for credential issuing in the cheqd ledger. And the killer feature is an opportunity to use coins, DID, credential issuing in one place.
But of course it's a question for Product Owners of Evernym and Cheqd.

Anyway, if we don't want to store a schema in the ledger in the future, it shouldn't affect this transaction.

architecture/adr-list/adr-002-identity-transactions.md
Toktar: Brent gave us the idea that schema and cred_def are resources, DDDoc simply describes a specific identifier.
We discussed this idea a lot earlier within the architectural team. Sergey K (as an Indy architect) and I do not recommend turning the schema and cred_def into a DiDDoc. These entities have a different class and a different unique set of fields for each resource.

And this file describes a ledger format of transaction. Theoretically, we can transform it on a client side

architecture/adr-list/adr-002-identity-transactions.md
Toktar: Started discussion in the previous thread.

architecture/adr-list/adr-002-identity-transactions.md
Toktar: I completely agree with this idea. In this way, we trim the original DiDDoc, but we keep all business cases, as well as make this interface more intuitive.

architecture/adr-list/adr-002-identity-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+## Decision
+
+### cheqd DID Method
+
+The `did:cheqd` method DID has four components that are concatenated to make a W3C DID specification conformant identifier. The components are:
+
+- **DID**: the hardcoded string `did:` to indicate the identifier is a DID
+- **`cheqd` DID method**: the hardcoded string `cheqd:` indicating that the identifier uses the `cheqd` DID method specification.
+- **Namespace**: A string that identifies the name of the primary cheqd ledger ("mainnet"), followed by a `:`. The namespace string may optionally have a secondary ledger name prefixed by a `:` following the primary name. If there is no secondary ledger element, the DID resides on the primary ledger ("mainnet"), else it resides on the secondary ledger. By convention, the primary is a production ledger while the secondary ledgers are non-production ledgers (e.g. testnet) associated with the primary ledger.
+- **Namespace Identifier**: A self-certified identifier unique to the given cheqd DID ledger namespace. To be self-certifying, the identifier must be derived from the initial `verkey` associated with the identifier.
```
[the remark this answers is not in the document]
Toktar: Fixed

architecture/adr-list/adr_002_identity_transactions.md
Toktar: This doc more about entities storing in the state that define fields list and write/read approach. Description for requests (transactions in the ledger) in this doc https://github.com/cheqd/cheqd-node/blob/main/docs/identity-api.md (needs to be updated)

architecture/adr-list/adr_002_identity_transactions.md
Toktar: Fixed

architecture/adr-list/adr_002_identity_transactions.md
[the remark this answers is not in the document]
Toktar: Added `alsoKnownAs` and `keyAgreement`

architecture/adr-list/adr_002_identity_transactions.md
```diff
@@ -44,103 +44,191 @@ _**Note**: Hyperledger Indy also contains other transaction types beyond the one
...
-**Note:** Fields `dest` and `owner` should have the same value.
 
-#### State format
+#### Verification Method
 
-`id -> {(alias, id, verkey), last_tx_hash, last_update_timestamp }`
+1. **`id` (string):**
 
-_DID transaction format:_
+   A string with format `<DIDDoc-id>#<key-alias`
```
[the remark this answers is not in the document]
Toktar: Fixed

architecture/adr-list/adr_002_identity_transactions.md
```diff
@@ -44,103 +44,191 @@ _**Note**: Hyperledger Indy also contains other transaction types beyond the one
...
 ### General structure of transaction requests
 
 All identity requests will have the following format:
 
 ```text
 {
     "data": { <request data for writing a transaction to the ledger> },
-    "owner": "GEzcdDLhCpGCYRHW82kjHd",
-    "signature": "49W5WP5jr7x1fZhtpAhHFbuUDqUYZ3AKht88gUjrz8TEJZr5MZUPjskpfBFdboLPZXKjbGjutoVascfKiMD5W7Ba",
+    "creators": [<identifier>, ...]
```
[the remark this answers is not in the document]
Toktar: We need to check a signature of request on the node side anyway. This field is redundant for DIDs, but remains useful for Schema and Cred Def until we added a controller to them. I think it makes sense to do this, although we rejected this option during the discussion of architecture.

Toktar:
**Topics for discussions:**
- reply protection
- make schema, cred_def in a format of DIDDoc 
- comparing with Cosmos Cash
- make `DIDDoc.authentication` and other list of keys just lists of key Ids.
- store fully qualified id for DID

**TODO:**
Add info about
- external refs
- transactions validation

architecture/adr-list/adr-002-identity-transactions.md
ankurdotb: A good example of this is available if you look up did:sov:WRfXPg8dantKVubE3HX8pw on https://dev.uniresolver.io/

You'll see here the services are defined as:
- agent: https://agents.danubeclouds.com/agent/WRfXPg8dantKVubE3HX8pw
- xdi: https://xdi03-at.danubeclouds.com/cl/+!:did:sov:WRfXPg8dantKVubE3HX8pw

Another example on ION is if you look up: did:ion:EiClkZMDxPKqC9c-umQfTkR8vvZ9JPhl_xLDI9Nfk38w5w where the services are:
- linkedin: linkedin.com/in/henry-tsai-6b884014
- github: github.com/thehenrytsai

architecture/adr-list/adr_002_identity_transactions.md
ankurdotb: @Toktar What are your thoughts on this?

architecture/adr-list/adr_002_identity_transactions.md
ankurdotb: To clarify, my point here was that we need to account for DIDDoc metadata, as Cosmos Cash implemented, like below:

```json
   "didDocumentMetadata":{
      "versionId":"4f0f8857ab36bdeee0ddb541ea7e7b9cb509d29e1103cc7def44d3d1e8220c22",
      "created":"2021-08-23T08:24:26.972761898Z",
      "updated":"2021-08-24T15:54:40.902858856Z"
   },
```

And also, this bit in their ADR is useful to note:

> In both cases, the target DID must exist on-chain and the signerAccount must exist as a verification method in a verification relationship of type authentication or it's key DID be listed as a DID controller. Additionally the DID document MUST NOT be in a deactivated status.
> 
> Each update operation MUST update the versionId and the updated field of the associated DID document metadata with the transaction hash and the block time respectively.

architecture/adr-list/adr-002-identity-transactions.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Decision
+
+### cheqd DID Method
+
+The `did:cheqd` method DID has four components that are concatenated to make a W3C DID specification conformant identifier. The components are:
+
+- **DID**: the hardcoded string `did:` to indicate the identifier is a DID
+- **`cheqd` DID method**: the hardcoded string `cheqd:` indicating that the identifier uses the `cheqd` DID method specification.
+- **Namespace**: A string that identifies the name of the primary cheqd ledger ("mainnet"), followed by a `:`. The namespace string may optionally have a secondary ledger name prefixed by a `:` following the primary name. If there is no secondary ledger element, the DID resides on the primary ledger ("mainnet"), else it resides on the secondary ledger. By convention, the primary is a production ledger while the secondary ledgers are non-production ledgers (e.g. testnet) associated with the primary ledger.
```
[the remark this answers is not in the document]
ankurdotb: @Toktar Would like your view on this. It's partly linked to storing full DIDs on ledger

architecture/adr-list/adr-002-identity-transactions.md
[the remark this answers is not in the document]
ankurdotb: DID Document metadata is a requirement of the DID Core specification. If we don't have them, then we don't have a DIDDoc. See: https://www.w3.org/TR/did-core/#did-document-metadata

In the Cosmos Cash implementation, they did all of the ones that were "should" in the DID Core spec and left the "may" ones, but ideally we should have all fields in the spec: https://github.com/allinbits/cosmos-cash/blob/main/docs/Explanation/ADR/adr-004-did.md#did-document-metadata

Finally, on deactivated status, it doesn't stop the DID from being read, it just ensures that there's a flag that it's no longer in use, and that no further update operations can be carried out on it. I don't see this as a business/legal risk.

architecture/adr-list/adr-002-identity-transactions.md
ankurdotb: It might impact the format of what's stored, in case the schema is an HTTP URL, like https://schema.org/docs/full.html

So if this is implement as a DIDDoc, it's more extensible.

architecture/adr-list/adr-002-identity-transactions.md
ankurdotb: I guess this is a broader conversation with Richard perhaps but the rest of the SSI ecosystem is moving towards DIDDoc to use this, so how much do we wanna future proof?

architecture/adr-list/adr-002-identity-transactions.md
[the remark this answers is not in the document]
ankurdotb: Okay sorry I looked at this in the actual changelog now and I see your point. Yes this metadata is different than the other one. Let's consider and discuss though whether this metadata is superseded by others.

ankurdotb (commented):
Will discuss during call today with @Toktar

architecture/adr-list/adr_002_identity_transactions.md
ankurdotb: @Toktar to check with Sergey K. DID Core spec I'm pretty sure needs fully qualified.

architecture/adr-list/adr-002-identity-transactions.md
ankurdotb: https://www.w3.org/TR/did-core/#services

architecture/adr-list/adr-002-identity-transactions.md
ankurdotb: did:cheqd:mainnet:identifier <- DIDDoc with JSON-LD description of the schema. Discussed with @Toktar, needs follow-up with Brent/Richard/Sergey K.

ankurdotb:
> **Topics for discussions:**
> 
> * reply protection
> * make schema, cred_def in a format of DIDDoc
> * comparing with Cosmos Cash
> * make `DIDDoc.authentication` and other list of keys just lists of key Ids.
> * store fully qualified id for DID
> 
> **TODO:** Add info about
> 
> * external refs
> * transactions validation

- Reply protection: Discussed and consider whether this is implemented by using the didDocumentMetadata fields
- Schema, creddef as DIDDocs: Needs more discussion with Richard/Brent/Sergey
- Cosmos Cash: Last time we checked their ADR, we were looking at using Indy format DIDs. Let's check, especially if the answer to above is different.
- DIDDoc.authentication: Let's check against which other DID Methods Evernym want to support
- Store fully qualified DID: Agreed
- External references: For external DID method references as well as non ledger references, it should be allowed on ledger because the client app should verify if these exist. So cheqd-node doesn't need to check, but connect.me and verity might need to do this.
- Tx validation: Reply protection and external references (as an example)

architecture/adr-list/adr-002-identity-transactions.md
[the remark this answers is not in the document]
brentzundel: you shouldn't need to store full DIDs on ledger. It is only the DID Doc that is returned when the DID is resolved that should have the full DID.

The identifier on the ledger is 1) already known to be a DID and 2) already on the ledger indicated by the namespace, so storage of those elements along with the namespace identifier seems redundant.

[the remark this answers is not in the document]
ankurdotb: @brentzundel To clarify my comment, I meant that the DIDDoc should have full DID. Right now, the ADR that's written suggests that it shouldn't. Agree with you on the rest.

Toktar:
Questions for the new Schema and Credential definition format:
- Cred_def has a `controller`. does it mean that it can be changed?
- Schema owner and `controller` field. The same question.
- Are we going to use fully qualified DID everywhere? If yes, is it ok that Schema DIDDoc and Schema entity have the same `id`?
- Is didDocumentMethadata for Schema DIDDoc is needed?
- Should it be a different key in the State for Schema DIDDoc?

ankurdotb (approved):

PR #153, opened 2021-10-04
Update node-ci.yml workflow and add installation tests [DEV-179] + [DEV-458]

VladimirWork:
This needs to be merged with **SQUASH**.

### DIFF
--- a/.github/workflows/node-ci.yml
+++ b/.github/workflows/node-ci.yml
@@ -101,7 +101,7 @@ jobs:
           cd test/networks/docker_compose
           ./gen_node_configs.sh
           ./run_docker.sh > /dev/null &
-          sleep 120
+          sleep 300
           docker ps
 
       - name: Add observer node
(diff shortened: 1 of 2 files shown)


.github/workflows/node-ci.yml
```diff
@@ -28,10 +62,17 @@ jobs:
...
-    name: Build cheqd-node and testnet Docker images
-    runs-on: ubuntu-latest
+  build-images-and-set-up-pool:
+    name: Build cheqd-node and testnet Docker images, install deb package and run Docker pool
+    runs-on: ubuntu-20.04 # exact version, not latest
+    needs: build
     steps:
+      - name: Download and install deb package
+        run: |
+          wget https://github.com/cheqd/cheqd-node/releases/download/v0.2.2/cheqd-node_0.2.2_amd64.deb > /dev/null
```
VladimirWork, line 72: Need to pass needed version here.

ankurdotb:
As discussed:

- Add Python Gitignore and remove cache files from being committed.
-  Tests in the tests/ folder, which might mean moving test/network underneath it OR since it's not really a test as such, moving it under docker/
- Change the hardcoded release/images to latest release instead. For Docker, Github Container Registry has a latest endpoint and the latest releases on releases should still resolve to correct wget

VladimirWork:
> * Tests in the tests/ folder, which might mean moving test/network underneath it OR since it's not really a test as such, moving it under docker/
@askolesov ^

askolesov:
I would keep networks under tests because they are used for manual testing.

PR #163, opened 2021-10-19
Add IBC parameters to the genesis ADR

askolesov:
Those values allow IBC connections from start. They can also be tweaked in runtime.


askolesov:
Updated

PR #174, opened 2021-10-31
Add swagger docs

Toktar:
TODO: add params description

### DIFF
--- a/docs/swagger.yaml
+++ b/docs/swagger.yaml
@@ -1,0 +1,215 @@
+swagger: '2.0'
+info:
+  title: Cosmos SDK - Legacy REST and gRPC Gateway docs
+  description: 'A REST interface for state queries, legacy transactions'
+  version: 1.0.0
+
+paths:
+  /getDid:
+    get:
+      tags:
+        - Query
+      summary: return DID
+      operationId: returnDid
+      description: |
+        Return DID Document from the ledger by DID
+      produces:
+        - application/json
+      parameters:
+        - in: query
+          name: id
+          description: id from DID Document. Fully qualified DID
+          required: true
+          type: string
+      responses:
+        200:
+          description: DID Document
+          schema:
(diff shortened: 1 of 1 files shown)


Toktar:
Ready for review

docs/swagger.yaml
```diff
@@ -0,0 +1,405 @@
...
+        description: A set of Service Endpoint maps
+        type: array
+        items:
+          $ref: '#/definitions/Service'
+      context:
+        description: A list of strings with links or JSONs for describing specifications that this DID Document is following to.
+        type: array
+        items:
+          type: string
+          example: "https://www.w3.org/ns/did/v1"
```
ankurdotb, line 222: Not sure if this is a Swagger doc issue only or in the actual code: the `context` section of a DID Document is right at the top, before everything else, and starts with `@context` before the block begins. If that's what we're doing, then this is a simple cosmetic fix in this file, otherwise we might need to check our implementation.

docs/swagger.yaml
```diff
@@ -0,0 +1,405 @@
...
```
ankurdotb, line 313: Same comment as the previous one on `context`

ankurdotb (changes requested):

docs/swagger.yaml
Toktar: Unfortunately, the field name in protobufs cannot start with `@`, but I think we can add it in DIDDoc resolver in JSON converter for passing the test suites.
Regarding fields order, it is not strictly defined in JSON, only in protobuff. In the proto and documentation, it is located at the end, since I followed the logic of specifying the required fields first, such as ID. Nevertheless, I wanted to separate the context from the other fields, so it was moved to the end.
In fact, we can move `context` to the top, if necessary.

ankurdotb:
@Toktar Swap the context field to top of the DIDDoc, before ID

ankurdotb (approved):

PR #176, opened 2021-10-31
Import and Export DID Genesis 

asmarty:


x/cheqd/keeper/did.go
```diff
@@ -91,3 +91,19 @@ func (k Keeper) HasDid(ctx sdk.Context, id string) bool {
 func GetDidIDBytes(id string) []byte {
 	return []byte(id)
 }
+
+// GetAllDid returns all did
+func (k Keeper) GetAllDid(ctx sdk.Context) (list []types.StateValue) {
```
askolesov, line 96: Iterator would be better here...

x/cheqd/types/genesis.go
```diff
@@ -1,21 +1,31 @@
...
+		DidList: []*StateValue{},
 	}
 }
 
 // Validate performs basic genesis state validation returning an error upon any
 // failure.
 func (gs GenesisState) Validate() error {
-	// this line is used by starport scaffolding # genesis/types/validate
-	// Check for duplicated ID in credDef
+	didIdMap := make(map[string]bool)
```
askolesov, line 15: Static validation should be called here in the future.

askolesov (approved):
To fix in future:
- We need an iterator instead of `get_all_dids`.
- Static validation should be called in genesis import.

PR #178, opened 2021-11-02
Save namespace into state

asmarty:


x/cheqd/types/genesis.go
```diff
@@ -27,5 +30,9 @@ func (gs GenesisState) Validate() error {
 		didIdMap[did.Id] = true
 	}
 
+	if len(gs.GetDidNamespace()) == 0 {
```
askolesov, line 33: We should accept empty namespace.

x/cheqd/keeper/msg_server_did.go
```diff
@@ -13,7 +13,7 @@ import (
 
 func (k msgServer) CreateDid(goCtx context.Context, msg *types.MsgCreateDid) (*types.MsgCreateDidResponse, error) {
 	ctx := sdk.UnwrapSDKContext(goCtx)
-	prefix := types.DidPrefix + ":" + types.DidMethod + ":" + ctx.ChainID() + ":"
+	prefix := types.DidPrefix + ":" + types.DidMethod + ":" + k.GetDidNamespace(ctx) + ":"
```
askolesov, line 16: And if the namespace is empty, handle it gracefully. I mean not put the second `:` in a row.

askolesov (changes requested):

Toktar (approved):

PR #186, opened 2021-11-09
Bump cosmos sdk to v0.44.3

askolesov:


ankurdotb (approved):

go.mod
```diff
@@ -12,9 +12,9 @@ require (
 	github.com/grpc-ecosystem/grpc-gateway v1.16.0
 	github.com/rakyll/statik v0.1.7
 	github.com/spf13/cast v1.3.1
-	github.com/spf13/cobra v1.1.3
+	github.com/spf13/cobra v1.2.1
 	github.com/spf13/pflag v1.0.5
-	github.com/tendermint/tendermint v0.34.12
+	github.com/tendermint/tendermint v0.34.14
 	github.com/tendermint/tm-db v0.6.4
 	google.golang.org/genproto v0.0.0-20210903162649-d08c68adba83
```
ankurdotb, line 19: Upstream is `google.golang.org/genproto v0.0.0-20210602131652-f16073e35f0c` in case we can change

ankurdotb (changes requested):
Added notes on a few more that could be bumped, this is a nice cleanup!

Does this remove the dependency (indirect) on the vulnerable versions of `gin-gonic` and `nyhoor/websocket`? Cosmos SDK upstream has an additional replace. https://github.com/cosmos/cosmos-sdk/blob/b75c29fc15d3320ec0c7596dbd7c787c48dccad8/go.mod#L67

go.mod
askolesov: Correct me if I'm wrong but 2021_09_03_162649 (current) is newer then 2021_06_02_131652 (you propose). The latest is 2021_11_04_193956, I suggest using it then.

Also, the more upgrades we do, the easier it is to break sth. That's why I updated only the necessary ones. (especially there are no breaking changes)

ankurdotb (approved):

PR #203, opened 2021-11-11
Create a genesis file for MainNet

Toktar:


persistent_chains/mainnet/genesis.json
```diff
@@ -0,0 +1,198 @@
...
+        "min_deposit": [
+          {
+            "denom": "ncheq",
+            "amount": "8000000000000"
+          }
+        ],
+        "max_deposit_period": "1210000s"
+      },
+      "voting_params": {
+        "voting_period": "1210000s"
```
Toktar, line 104: @ankurdotb Should we increase this value?

mircea-c: > @ankurdotb Should we increase this value?

Really depends on the validator set (who controls the vote). There are a number of configuration options that can be changed through governance and things like recovering an expired IBC client. Would you be willing to wait 2-3 weeks for something like that?

Toktar: Sorry, my bad.  I meant decreasing because the current situation in TestNet shows that 2 weeks is too long.

ankurdotb:
@mircea-c We updated the genesis file to reduce some of those parameters such as voting periods etc.

ankurdotb (approved):

PR #209, opened 2021-11-12
Add chown for directories in .deb

anikitinDSR:


askolesov (approved):
Is there a reason why we create `data` and `config` directories? IMO we should either create and use them or we should allow `init` to do it.

ankurdotb (approved):

PR #215, opened 2021-11-13
Updated links in SUMMARY.md

Tweeddalex:


ankurdotb:
Will edit this separately to fix instead given the suggested edits are not the target state we want

PR #220, opened 2021-11-15
fix: Referencing last version

italobb:
The tutorial was failing because version v0.2.3 doesn't have a `configure` command. Updated to version v0.2.4 to fix the issue.


ankurdotb:
Hey Italo, this is now superseded by the new version of the docs

PR #237, opened 2021-11-30
Add docker-compose for running one node

anikitinDSR:

### DIFF
--- a/persistent_chains/explorer/bigdipperv1-mainnet-config.json
+++ b/persistent_chains/explorer/bigdipperv1-mainnet-config.json
@@ -13,7 +13,7 @@
         "bech32PrefixConsAddr": "cheqdvalcons",
         "bech32PrefixConsPub": "cheqdvalconspub",
         "bondDenom": "ncheq",
-        "powerReduction": 1000000000,
+        "powerReduction": 1000000,
         "coins": [
             {
                 "denom": "ncheq",
--- a/persistent_chains/explorer/bigdipperv1-testnet-config.json
+++ b/persistent_chains/explorer/bigdipperv1-testnet-config.json
@@ -13,7 +13,7 @@
         "bech32PrefixConsAddr": "cheqdvalcons",
         "bech32PrefixConsPub": "cheqdvalconspub",
         "bondDenom": "ncheq",
-        "powerReduction": 1000000000,
+        "powerReduction": 1000000,
         "coins": [
             {
                 "denom": "ncheq",


docker/cheqd_node/mainnet_launch/.env
```diff
@@ -0,0 +1,10 @@
...aWduZWRfcGVyX3dpbmRvdyI6ICIwLjUwMDAwMDAwMDAwMDAwMDAwMCIsCiAgICAgICAgImRvd250aW1lX2phaWxfZHVyYXRpb24iOiAiNjAwcyIsCiAgICAgICAgInNsYXNoX2ZyYWN0aW9uX2RvdWJsZV9zaWduIjogIjAuMDUwMDAwMDAwMDAwMDAwMDAwIiwKICAgICAgICAic2xhc2hfZnJhY3Rpb25fZG93bnRpbWUiOiAiMC4wMTAwMDAwMDAwMDAwMDAwMDAiCiAgICAgIH0sCiAgICAgICJzaWduaW5nX2luZm9zIjogW10sCiAgICAgICJtaXNzZWRfYmxvY2tzIjogW10KICAgIH0sCiAgICAic3Rha2luZyI6IHsKICAgICAgInBhcmFtcyI6IHsKICAgICAgICAidW5ib25kaW5nX3RpbWUiOiAiMTIxMDAwMHMiLAogICAgICAgICJtYXhfdmFsaWRhdG9ycyI6IDEyNSwKICAgICAgICAibWF4X2VudHJpZXMiOiA3LAogICAgICAgICJoaXN0b3JpY2FsX2VudHJpZXMiOiAxMDAwMCwKICAgICAgICAiYm9uZF9kZW5vbSI6ICJuY2hlcSIKICAgICAgfSwKICAgICAgImxhc3RfdG90YWxfcG93ZXIiOiAiMCIsCiAgICAgICJsYXN0X3ZhbGlkYXRvcl9wb3dlcnMiOiBbXSwKICAgICAgInZhbGlkYXRvcnMiOiBbXSwKICAgICAgImRlbGVnYXRpb25zIjogW10sCiAgICAgICJ1bmJvbmRpbmdfZGVsZWdhdGlvbnMiOiBbXSwKICAgICAgInJlZGVsZWdhdGlvbnMiOiBbXSwKICAgICAgImV4cG9ydGVkIjogZmFsc2UKICAgIH0sCiAgICAidHJhbnNmZXIiOiB7CiAgICAgICJwb3J0X2lkIjogInRyYW5zZmVyIiwKICAgICAgImRlbm9tX3RyYWNlcyI6IFtdLAogICAgICAicGFyYW1zIjogewogICAgICAgICJzZW5kX2VuYWJsZWQiOiB0cnVlLAogICAgICAgICJyZWNlaXZlX2VuYWJsZWQiOiB0cnVlCiAgICAgIH0KICAgIH0sCiAgICAidXBncmFkZSI6IHt9LAogICAgInZlc3RpbmciOiB7fQogIH0KfQo=
```
ankurdotb, line 1: Ah so I wouldn't want to pass it this way. Basically the better way to pass a file is to use a -v or volume argument to the Docker container, specifying where persistent storage like genesis file or indeed other things like priv validator key, node keys, etc are kept. See this example: https://www.docker.com/blog/how-to-use-the-official-nginx-docker-image/#:~:text=%24%20docker%20run%20%2Dit%20%2D%2Drm%20%2Dd%20%2Dp%208080%3A80%20%2D%2Dname%20web%20%2Dv%20~/site%2Dcontent%3A/usr/share/nginx/html%20nginx

It provides a clean and consistent way of specifying config files, data files etc that are consistent.

ankurdotb, line 1: Ah so I wouldn't want to pass it this way. Basically the better way to pass a file is to use a -v or volume argument to the Docker container, and then copying it over to the path needed. This way anyone can download the latest genesis file, seed file etc and pass it to the Docker container as a volume argumment. See this example: https://www.docker.com/blog/how-to-use-the-official-nginx-docker-image/#:~:text=%24%20docker%20run%20%2Dit%20%2D%2Drm%20%2Dd%20%2Dp%208080%3A80%20%2D%2Dname%20web%20%2Dv%20~/site%2Dcontent%3A/usr/share/nginx/html%20nginx

It provides a clean and consistent way of specifying config files that is consistent.

docker/cheqd_node/mainnet_launch/.env
```diff
@@ -0,0 +1,10 @@
...
+
+P2P_SEEDS=258a9bfb822637bfca87daaab6181c10e7fd0910@seed1.eu.cheqd.net:26656,f565ff792b20977face9817df6acb268d41d4092@seed2.eu.cheqd.net:26656,388947cc7d901c5c06fedc4c26751634564d68e6@seed3.eu.cheqd.net:26656,9b30307a2a2819790d68c04bb62f5cf4028f447e@seed1.ap.cheqd.net:26656,debcb3fa7d40e681d98bcc7d22278fd58a34b73a@144.76.183.180:1234,abd4be300be882ae9a69ab0959260afe8871f7a6@165.232.167.156:26656,fd17fe46e8b69bfa006b3fba53cfc9df2b8f9512@161.35.177.151:26656
+
+RPC_LADDR=tcp://0.0.0.0:26657
+
+# The next parameters should be changed by user
+NODE_MONIKER=<name-of-node>
```
ankurdotb, line 8: All environment variables can be provided in a single shot under --env-file. This can be potentially very useful as well since we could have a single file with all values that people need to edit that can be dropped in, e.g., everything that cheqd-noded configure now supports.

docker/cheqd_node/mainnet_launch/.env
```diff
@@ -0,0 +1,10 @@
...
+
+OUT_MOUNT_POINT=/path/to/data/on/the/host
```
ankurdotb, line 10: Consider mount points (used above) vs using volumes in general. Volumes will ensure that there's consistent permissions and ownership that can be applied to the contents. See this for a comparison: https://docs.docker.com/engine/reference/commandline/service_create/#add-bind-mounts-volumes-or-memory-filesystems

docs/setup-and-configure/docker-install.md
```diff
@@ -69,3 +70,19 @@ Usage:
...
 ```
+
+### Mainnet launch
+
+For simple running node which can be connected to `cheqd-mainnet-1` the next steps can be useful:
+* Go to `docker/cheqd_node/mainnet`
+* Edit parameters `NODE_MONIKER` and `OUT_MOUNT_POINT` in the `.env` file and run:
+
+```
+docker-compose up
```
ankurdotb, line 81: I'd change this to explicitly specify the env file used by Compose. Otherwise if there are multiple files for env (this is one way, btw, we can ourselves have multiple env files for different types of deployments) https://docs.docker.com/engine/reference/commandline/service_create/#add-bind-mounts-volumes-or-memory-filesystems

ankurdotb (changes requested):
Overall general comment: what do other Cosmos projects do?

docker/cheqd_node/node-runner.sh
```diff
@@ -5,16 +5,33 @@
...
 
 CHEQD_ROOT_DIR="$HOME/.cheqdnode"
+CHEQD_COMMON_DIR="$HOME/.common"
 
 # Init node config directory
 if [ ! -d "${CHEQD_ROOT_DIR}/config" ]
 then
     echo "Node config not found. Initializing."
-    cheqd-noded init $NODE_MONIKER --home 
+    cheqd-noded init $NODE_MONIKER --home ${CHEQD_ROOT_DIR}
```
ankurdotb, line 14: This should map to the volume path shouldn't it? Otherwise, the volume path and this path could likely in in different places?

docker/cheqd_node/mainnet_launch/docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+
+services:
+  node:
+    image: cheqd-node
+    ports:
+      - "26656:26656" # p2p
+      - "26657:26657" # rpc
+    volumes:
+      - cheqd_data:/cheqd
+      - ../../../persistent_chains/mainnet/:/cheqd/.common
```
ankurdotb, line 11: Not sure of this path..., why does it need to look a few level up?

docker/cheqd_node/node-runner.sh
```diff
@@ -5,16 +5,33 @@
...
+    echo "There is no genesis.json in ${CHEQD_COMMON_DIR} directory. Variable value GENESIS will be used"
+fi 
+
+if [ -e "${CHEQD_COMMON_DIR}/seeds.txt" ]
+then
+    echo "Use seeds.txt from ${CHEQD_COMMON_DIR} directory"
+    P2P_SEEDS=$(cat ${CHEQD_COMMON_DIR}/seeds.txt)
+else
+    echo "There is no seeds.txt in ${CHEQD_COMMON_DIR} directory. Variable value P2P_SEEDS will be used"
+fi
```
ankurdotb, line 33: Not sure where this is set/passed in. Maybe I'm just tired

docker/cheqd_node/node-runner.sh
```diff
@@ -36,7 +53,7 @@ if [[ ! -z ${P2P_SEND_RATE+x} ]]; then cheqd-noded configure p2p send-rate ${P2P
 set - # Disable ditailed command logging
 
 echo "Updating genesis"
-echo "$GENESIS" | base64 --decode > $CHEQD_ROOT_DIR/config/genesis.json
+echo "$GENESIS" > $CHEQD_ROOT_DIR/config/genesis.json
```
ankurdotb, line 56: If the common DIR is being used to pass in the genesis.json, there should be no need to read the genesis file, create an environment variable, and then go set it again back to a file. Either it's new node+volume in which case pass it in using tmpfs, or it's an existing volume in which the genesis file is already there.

Basically, let the user decide whether to pass in config use the compose file. If there's no genesis file, let the container panic and crash.

ankurdotb (changes requested):

docker/cheqd_node/node-runner.sh
anikitinDSR: Well, for now it doesn't work, cause `--home` requires a value after. Actually, for now I'm going to get rid of `--home` parameter at all.

docker/cheqd_node/mainnet_launch/docker-compose.yml
anikitinDSR: I just don't want to copy/paste genesis and seeds into `mainnet_launch` directory. It's default behaviour and acan be changed by user of course.

docker/cheqd_node/node-runner.sh
anikitinDSR: It's passed by mounting another volume in the `docker-compose`, especially `- ../../../persistent_chains/mainnet/:/cheqd/.common`

docker/cheqd_node/node-runner.sh
anikitinDSR: Yup, I was choosing between getting rid of it at all and keep the ability to pass GENESIS as an environment parameter just for running docker container without composing.

ankurdotb (changes requested):

docker/cheqd_node/mainnet_launch/docker-compose.yml
ankurdotb: After today's discussion: this could be a variable in the .env file

jay-dee7:
@ankurdotb @ankurdotb for some weird reason, the p2p seeds part in the node-configure command doesn't seem to work, i tested on my mac and it worked but apparently not with the containers.
Also I tried so much to make command+entrypoint work in compose work but it just doesn't work. To my surprise, compose only carries forward one command from command and entrypoint and not both (like it does on normal docker run)
Here's a supporting issue for this one

https://github.com/docker/compose/issues/3140

PR #240, opened 2021-12-02
Added Makefile build, clean

Eengineer1:


Makefile
```diff
@@ -0,0 +1,123 @@
...
+ifeq (rocksdb,$(findstring rocksdb,$(COSMOS_BUILD_OPTIONS)))
+  CGO_ENABLED=1
+  BUILD_TAGS += rocksdb
+  ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=rocksdb
+endif
+# handle boltdb
+ifeq (boltdb,$(findstring boltdb,$(COSMOS_BUILD_OPTIONS)))
+  BUILD_TAGS += boltdb
+  ldflags += -X github.com/cosmos/cosmos-sdk/types.DBBackend=boltdb
+endif
```
ankurdotb, line 74: To be honest I don't think we'll be using any of the other DB backends, we'll likely use either CLevelDB going forward or GoLevelDB. I'd much rather skip this to not set the expectation that we support these backends (someone who really wants to can go figure this edge case themselves)

Makefile
```diff
@@ -0,0 +1,123 @@
+PACKAGES_NOSIMULATION=$(shell go list ./... | grep -v '/simulation')
```
anikitinDSR, line 1: Do we need this parameters?
`PACKAGES_NOSIMULATION`
`PACKAGE_SIMTEST`

Makefile
```diff
@@ -0,0 +1,123 @@
...
+BINDIR ?= $(GOPATH)/bin
+BUILDDIR ?= $(CURDIR)/build
+
+export GO111MODULE = on
+
+# process build tags
+
+build_tags = netgo
+ifeq ($(LEDGER_ENABLED),true)
+  ifeq ($(OS),Windows_NT)
```
anikitinDSR, line 16: I think, that we need to keep only Linux as our target OS. In other cases, people can think that our binary can work on other OS.

Makefile
```diff
@@ -0,0 +1,123 @@
...
+
+whitespace :=
+whitespace += $(whitespace)
+comma := ,
+build_tags_comma_sep := $(subst $(whitespace),$(comma),$(build_tags))
+
+# process linker flags
+
+ldflags = -X github.com/cosmos/cosmos-sdk/version.Name=cheqd-node \
+		  -X github.com/cosmos/cosmos-sdk/version.AppName=cheqd-noded \
```
askolesov, line 50: Let's move Name and AppName to constants too. Btw, what is the difference?

Makefile
```diff
@@ -0,0 +1,123 @@
...
+  BUILD_FLAGS += -gcflags "all=-N -l"
+endif
+
+all: build
+
+BUILD_TARGETS := build install
+
+build: BUILD_ARGS=-o $(BUILDDIR)/
+
+build-linux:
```
askolesov, line 102: What is the difference between build and build-linux?

Makefile
```diff
@@ -0,0 +1,123 @@
...
+	GOOS=linux GOARCH=$(if $(findstring aarch64,$(shell uname -m)) || $(findstring arm64,$(shell uname -m)),arm64,amd64) LEDGER_ENABLED=false $(MAKE) build
```
askolesov, line 103: There was LEDGER_ENABLED constant introduced, but we use `false` here

askolesov (changes requested):

Makefile
Eengineer1: Sure, removed to avoid confusion.

Makefile
Eengineer1: No, they are leftovers from a previous version I was working on. Removed.

Makefile
Eengineer1: Moved to constants, too.
I think the difference here, if not mistaken, is an AppName for non-separate builds for server and cli.
In case of separate builds, we have to define ServerName and ClientName.
The Name is included in both cases.

Makefile
Eengineer1: Removed as it was used only for testing, no substantial difference here.

Makefile
Eengineer1: Same as above.

Makefile
askolesov: Things were easier: https://github.com/cosmos/cosmos-sdk/blob/master/version/version.go#L27

askolesov: @ankurdotb We have two constants for `application's name` and `application binary name`.
- Binary name: Jacob suggested renaming cheqd-noded to cheqd. Do we consider changing it?
- App name: What should be there?

ankurdotb (commented):
Will defer to @Toktar, @askolesov or @anikitinDSR but what about the cheqd proto definitions?

Makefile
ankurdotb: No let's keep it cheqd-noded please. Since the name is cheqd and the default is to add a "d" at the end, I don't want to be "cheqdd".

askolesov (approved):

Toktar (approved):

.github/workflows/node-ci.yml
```diff
@@ -50,24 +50,43 @@ jobs:
...
+            github.com/gogo/protobuf/protoc-gen-gogofast@v${GOGO_PROTOBUF_VERSION} \
+            github.com/gogo/protobuf/protoc-gen-gogofaster@v${GOGO_PROTOBUF_VERSION} \
+            github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway@v${GRPC_GATEWAY_VERSION} \
+            github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger@v${GRPC_GATEWAY_VERSION} \
+            github.com/regen-network/cosmos-proto/protoc-gen-gocosmos@latest
+
+      - name: Install buf
+        env:
+          PREFIX: "/usr/local"
+          VERSION: "1.0.0-rc8"
```
anikitinDSR, line 72: Why do we use here RC version?

Makefile
anikitinDSR: And again, why do we use here Windows instructions too?

.github/workflows/node-ci.yml
```diff
@@ -50,24 +50,43 @@ jobs:
...
+        run: |
+          curl -sSL "https://github.com/bufbuild/buf/releases/download/v${VERSION}/buf-$(uname -s)-$(uname -m).tar.gz" | \
+          sudo tar -xvzf - -C "${PREFIX}" --strip-components 1
 
       - uses: actions/checkout@v2
         with:
-          fetch-depth: 0 # FIXME: Starport requires full repository
+          fetch-depth: 0 # Required to fetch version
```
anikitinDSR, line 79: Should we get back fetch-depth parameter because we are getting rid of starport?

anikitinDSR (approved):

PR #242, opened 2021-12-06
Fix identity authentication

askolesov:


x/cheqd/types/keys.go
```diff
@@ -25,6 +25,7 @@ func KeyPrefix(p string) []byte {
 const (
 	DidKey      = "did:"
 	DidCountKey = "did-count:"
+	// DidNamespaceKey FIXME: Should be `did-namespace:`.
+	// Networks was started with `testnet` value so we need a migration now.
+	DidNamespaceKey = "testnet"
```
ankurdotb, line 30: This looks like a regression?

ankurdotb (commented):
Will leave this to @Toktar and @anikitinDSR to look in detail

x/cheqd/types/keys.go
askolesov: No, a wrong constant was used to take this value, I fixed that and made the fact that we used the wrong value more explicit

Toktar (approved):

tests/cli/tests/test_rotate_key.sh
```diff
@@ -0,0 +1,110 @@
+#!/bin/bash
+
+set -euox pipefail
+
+# sed in macos requires extra argument
+
+sed_extension=''
```
anikitinDSR, line 7: Is it still needed?

PR #255, opened 2021-12-15
Fix codecs

askolesov:
Fixed cheqd types registration.
Fixed proto packing/unpacking.

These issues caused panic during nymList export to genesis.


x/cheqd/types/keys.go
```diff
@@ -27,5 +27,5 @@ const (
 	DidCountKey = "did-count:"
 	// DidNamespaceKey FIXME: Should be `did-namespace:`.
 	// Networks was started with `testnet` value so we need a migration now.
-	DidNamespaceKey = "testnet"
+	DidNamespaceKey = "testnettestnet"
```
ankurdotb, line 30: Is this correct?

ankurdotb (commented):
One comment left behind regarding something that seems wrong on namespace. And just to confirm, this is overridden by genesis file namespace if it's present correct?

x/cheqd/types/keys.go
askolesov: Yep, it compensates this fix: https://github.com/cheqd/cheqd-node/pull/255/files#diff-abdef39a78c0c284466e9bf9b1c8510681273c75201228b173a6a6a1a466d989L11

askolesov:
This PR seems to be self-contained. Static validation will be implemented in another one.

x/cheqd/types/keys.go
ankurdotb: Maybe I'm being dense but explain me one last time tomorrow whether we still need this, given that both testnet and mainnet have this now set via genesis to "testnet" / "mainnet". If I understand correctly, this is internally stored as "testnettestnet" and "mainnetmainnet" but not exposed to the user. Fine. But given that both testnet and mainnet will be running exactly the same version of the software in the future, how will this validation account for that?

ankurdotb (commented):
Needs one final discussion tomorrow before closing out

askolesov:
Can we merge this now? I've created a ticket to migrate `did-namespace` key name: https://app.clickup.com/t/6600954/DEV-804.

ankurdotb (approved):

PR #264, opened 2021-12-23
Improve Github Actions by adding super linter

askolesov:


.github/workflows/lint.yml
```diff
@@ -0,0 +1,61 @@
+name: Lint
+
+on: [push]
```
ankurdotb, line 3: Do we want to break this up like earlier, so that say if I push Markdown files, only Markdown linting is done? For the rest, I'm okay with everything being executed in one go. Otherwise, I'd much rather there is a distinction between Markdown vs non-Markdown linting (which is usually code or code adjacent). The reason is that otherwise if I'm editing docs, the CI pipeline runs for a very long time and blocks merging.

The other reason is on our Github plan we have a limited number of free execution minutes and if we're running very long Actions sequences on every push, it's very wasteful in terms of minutes executed. So being a bit more intelligent on what Actions are executed on what push will help us optimise that.

.github/linters/.markdown-lint.yml
```diff
@@ -4,11 +4,11 @@ default: false # includes/excludes all rules by default
 MD001: true
 
 # Heading style <https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md#md003>
-MD003: true
+MD003:
```
ankurdotb, line 7: If I understand this correctly, the `default: false` turns off execution by default but if there's a `style` specified it turns the rule execution on without needing an explicit `true` flag. Is that correct?

If yes, can you please copy across these same config changes over to https://github.com/cheqd/.github/tree/main/workflow-templates so that we have it available consistently across all repos to copy from?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,61 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Super-Linter
+      uses: github/super-linter/slim@v4
+      env:
+        VALIDATE_GITLEAKS: true
+        VALIDATE_MARKDOWN: true
```
ankurdotb, line 45: Does this pick up / use the current config for Markdown? I'd ideally want to turn this to false and execute/handle Markdown separately. What I'm thinking of is also that Markdown linting is needed in completely non-code repos such as https://github.com/cheqd/cheqd-governance and I want to keep it consistent across all repos (there's no point running super linter for that repo)

.github/workflows/lint.yml
```diff
@@ -0,0 +1,61 @@
...
+    name: Run super linter
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+
+    - name: Super-Linter
+      uses: github/super-linter/slim@v4
+      env:
+        VALIDATE_GITLEAKS: true
```
ankurdotb, line 44: What does this do? We have Git secret scanning set up through a separate app.

ankurdotb (changes requested):

ankurdotb (approved):

PR #270, opened 2022-01-20
Get rid of circle dependencies from vdrtools

anikitinDSR:


ankurdotb:
Task linked: [DEV-720 Get rid of circular dependency between VDR and node](https://app.clickup.com/t/6600954/DEV-720)

Toktar (approved):

Toktar:
Before merging we need to fix `Test / Run cosmos and identity tests using CLI (push)` pipeline

PR #276, opened 2022-02-01
[DEV-826] Additional txns before upgrade

anikitinDSR:


tests/upgrade/common.sh
```diff
@@ -58,13 +59,95 @@ function docker_compose_down () {
...
 function clean_env () {
     rm -rf node_configs
+    rm -f $FNAME_TXHASHES
 }
 
 # Run command using local generated keys from node_configs/client
-function local_client_exec () {
-    cheqd_noded_docker "$@" --home node_configs/client/.cheqdnode/ --keyring-backend test
+function local_client_tx () {
+    result="$(cheqd_noded_docker $@ --home node_configs/client/.cheqdnode/ --keyring-backend test)"
```
Toktar, line 67: Could you please not to use a global variable `result` if it's possible?

Toktar (changes requested):

tests/upgrade/common.sh
```diff
@@ -58,13 +59,95 @@ function docker_compose_down () {
...
+        --ver-key "${ALICE_VER_PRIV_BASE_64}" \
+        --from operator0 \
+        --gas-prices "25ncheq" \
+        --chain-id $CHAIN_ID \
+        --output json \
+        -y
+
+
+    txhash=$(echo $result | jq ".txhash" | tr -d '"')
+    echo $txhash >> $FNAME_TXHASHES
```
Toktar, line 138: Could you please check transactions using query? TxHash is a good way too, let's keep it. But I'm not sure about TxHash after pool upgrade with migration

tests/upgrade/upgrade_and_check.sh
```diff
@@ -22,13 +22,21 @@ docker_compose_up "$CHEQD_IMAGE_TO" $(pwd)
...
+# Check that token transaction exists after upgrade too
+check_tx_hashes
+
+# Check balances after token sending
+check_balance
+
+# Check that did written before upgrade stil exist
+check_did
+
 # Stop docker compose
```
Toktar, line 40: Could you please write and check bank end did transactions after upgrade?

Toktar (changes requested):

Toktar (approved):

PR #280, opened 2022-02-11
Further docker, tests improvements [DEV 841]

askolesov:


ankurdotb:
Task linked: [DEV-841 Docker Compose + Actions pipeline improvements](https://app.clickup.com/t/6600954/DEV-841)

.github/workflows/release.yml
```diff
@@ -210,14 +182,6 @@ jobs:
...
 
-      - name: Download testnet image
-        uses: actions/download-artifact@v2
-        with:
-          name: cheqd-testnet-image.tar
-
-      - name: Load testnet image
-        run: docker load -i cheqd-testnet-image.tar
-
       - name: Login to docker
```
ankurdotb, line 185: Use the Github Actions recommended way using the specific actions and secret.GITHUB_TOKEN instead, as this is safer than using a PAT: https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages

docker/entrypoint.sh
```diff
@@ -0,0 +1,58 @@
...
+set -euo pipefail
+
+# within the container, $HOME=/home/cheqd
+CHEQD_ROOT_DIR="$HOME/.cheqdnode"
+
+# Init node config directory
+if [ ! -d "${CHEQD_ROOT_DIR}/config" ]
+then
+    echo "Node config not found. Initializing."
+    cheqd-noded init "moniker-placeholder"
```
ankurdotb, line 14: Just to confirm, is the value of `CHEQD_NODED_MONIKER` in `container.env` used as the actual moniker, if it has been set?

docker/localnet/container-env/observer-0.env
```diff
@@ -0,0 +1,11 @@
+# Environment variables used by the container
+
+# General
+CHEQD_NODED_MONIKER="observer-0"
+CHEQD_NODED_RPC_LADDR="tcp://0.0.0.0:26657"
```
ankurdotb, line 5: Don't you also need to set P2P_LADDR (or is that by default)?

docker/localnet/container-env/seed-0.env
```diff
@@ -0,0 +1,11 @@
+# Environment variables used by the container
+
+# General
```
ankurdotb, line 3: Since this is a seed, you're going to need P2P_SEED_MODE, otherwise this isn't a seed

.github/workflows/test.yml
```diff
@@ -280,50 +246,51 @@ jobs:
...
+        uses: actions/checkout@v2
 
-      - name: Set up 4 validators + 2 observers node docker pool for local test run # Replace with testnet image
-        working-directory: ./tests/networks/docker-compose-localnet
+      - name: Set up 4 validators + 2 observers node docker pool
+        working-directory: ./docker/localnet
         run: |
           bash gen-network-config.sh
-          docker-compose up -d
+          CHEQD_NODE_IMAGE=cheqd-node CHEQD_NODE_VERSION=latest docker compose up -d
```
ankurdotb, line 261: If you want to call/reference the file docker-compose.env for any variables you need in Docker Compose file itself (as opposed to within the container), the correct syntax for this is `docker compose --env-file <filepath to docker-compose.env> up -d`. You don't need to do it inline like this. 

BTW, this is what I meant by the fact that the `env_file` in a Docker Compose file is entirely different than the `env-file` flag for Docker Compose CLI. They don't make it easy!

ankurdotb (changes requested):
Thanks for this @askolesov, I know this is a huge PR. Please take a look at the comments highlighted and take a look at what you agree with / to fix.

A few generic comments (that applies to multiple nodes)

- Set fast sync version to v0 (if not already done in code)
- Set minimum gas price (to be equivalent to production networks)
- Try setting create empty blocks to false (if it isn’t already by default)

.github/workflows/test.yml
askolesov: Env file isn't supported by `docker compose` (with space) subcommand, `docker-compose` standalone utility is outdated on GitHub runners so I did that. So I see options:
* Update `docker-compose` manually and use it
Oh, I see only one option :)

.github/workflows/release.yml
askolesov: Yeah... Wanted to do it too. Finally managed to configure package permissions.

docker/entrypoint.sh
askolesov: Yes

docker/localnet/container-env/observer-0.env
askolesov: For p2p it's 0.0.0.0 by default while for RPC it's limited to localhost

docker/localnet/docker-compose.env
```diff
@@ -0,0 +1,12 @@
...
+###############################################################
+###             REQUIRED: Node/network selection           ###
+###############################################################
+# Define cheqd-noded image source. Change if using your own build.
+CHEQD_NODE_IMAGE="ghcr.io/cheqd/cheqd-node"
+
+# Define cheqd-noded software release version
+# Current MAINNET recommended version: v0.3.1
+# Current TESTNET recommended version: v0.3.1
+CHEQD_NODE_VERSION="0.3.1"
```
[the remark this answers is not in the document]
askolesov: > Use this env file to change variables you need within the Docker Compose file.
It's already so.

> Hard-code the published ports for each node.
Already done.

> Then, distinguish the nodes by setting different local IP addresses instead. Docker Compose in bridge mode can be used to define a local network, IP range etc. The advantage is for example if we ever wanted to set up a localnet on AWS, other provides etc where everything isn't just on one machine but across different machines, this can be easily changed by changing the "network" definition. See example in Tendermint repo: https://github.com/tendermint/tendermint/blob/master/docker-compose.yml

Everything is already as you described, even better than in Tendermint.
* Docker-compose creates a bridge network by default
* Instead of custom IPs/network masks we use DNS records that correspond to service names i.e. `observer-0`, 'validator-2', and so on.

askolesov:
@ankurdotb Thanks for the review! Happy to discuss the comments above in huddle to close this PR faster if it's OK for you.

>     * Set fast sync version to v0 (if not already done in code)
It's `v0` by default.

>     * Set minimum gas price (to be equivalent to production networks)
Good point, fixed.

>     * Try setting create empty blocks to false (if it isn’t already by default)
Done.

cmd/cheqd-noded/cmd/debug_extensions.go
```diff
@@ -42,7 +42,6 @@ func ed25519RandomCmd() *cobra.Command {
 
 			keyInfo := struct {
 				PubKeyBase64    	string `json:"pub_key_base_64"`
-				PubKeyMultibase58   string `json:"pub_key_multibase_58"`
```
ankurdotb, line 45: Isn't this used in our cheqd CLI identity instructions?

.github/workflows/test.yml
ankurdotb: Hmmm it's worth checking if the Docker-compose failure is because GitHub doesn't support it or if the Actions fails due to a similar reason. https://github.community/t/how-to-use-docker-compose-with-github-actions/16850/2

I do agree about needing to override certain commands. I guess the question is...the test pipeline tests against a localnet, correct? The persistent chain image can have a fixed version number, pulled from GHCR but for the test pipeline, I would image the version number and so can be declared as environment variables in the command. Basically there's a lot besides the node moniker and node version that is declared in the env file, that might be useful to actually take from the env file. What do you think?

ankurdotb (commented):
Final few discussions to close out

cmd/cheqd-noded/cmd/debug_extensions.go
askolesov: No, another command is used to transcode `base64` -> `multibase58`. This is the empty field that we saw during our session.

.github/workflows/test.yml
askolesov: > Hmmm it's worth checking if the Docker-compose failure is because GitHub doesn't support it or if the Actions fails due to a similar reason. https://github.community/t/how-to-use-docker-compose-with-github-actions/16850/2

We could use `--env-file` but we have to override all values in it: image name and image version. So imo there is no need for `--env-file`. Would you agree?

> I do agree about needing to override certain commands. I guess the question is...the test pipeline tests against a localnet, correct?

Yes

> ...but for the test pipeline, I would image the version number and so can be declared as environment variables in the command.

Here they are:

```
CHEQD_NODE_IMAGE=cheqd-node CHEQD_NODE_VERSION=latest docker compose up -d
```

There is no flag like `--env` in compose.

> Basically there's a lot besides the node moniker and node version that is declared in the env file, that might be useful to actually take from the env file. What do you think?

Node monikers are container-related so they are in container env files like [this one](https://github.com/cheqd/cheqd-node/pull/280/files/b743a7aa870b3c981f5864b1ebd3965de0ffff67#diff-730343d940a9be6d13468423951c9d5e00514286773a0ce5b1304d5342c40f45).

On the other hand image name and version are compose related parameters so they are in [docker-compose.env](https://github.com/cheqd/cheqd-node/pull/280/files/b743a7aa870b3c981f5864b1ebd3965de0ffff67#diff-3d4036a5f8ce5d689d2fb94ab18543fb255922399483d303b69968388c5de7e2).

In the current scenario we override both image and version so there is no need for `--env-file` flag in my opinion.

ankurdotb (approved):

PR #285, opened 2022-02-18
DEV-1079: Split validation into basic and dynamic

askolesov:


ankurdotb:
Task linked: [DEV-705 [performance] Split validation into basic and dynamic.](https://app.clickup.com/t/6600954/DEV-705)

x/cheqd/types/stateValue.go
```diff
@@ -24,7 +24,7 @@ func NewStateValue(data StateValueData, metadata *Metadata) (*StateValue, error)
 	return &StateValue{Data: any, Metadata: metadata}, nil
 }
 
-func NewMetadata(ctx sdk.Context) Metadata {
+func NewMetadataFromContext(ctx sdk.Context) Metadata {
```
ankurdotb, line 27: Our DID specification states the versionId is the transaction hash, not a base64 encoded version of the tx hash. I'm more inclined to keep it as just the tx hash, for instance, because that can easily be plugged into a block explorer to pull up that transaction. Do you agree?

ankurdotb (changes requested):
Discuss and check versionId deviation from
ADR. I know that when we launched, this wasn't correctly being set to the tx hash, and I would expect that as part of this validation, we validate it against the correct format that makes sense.

x/cheqd/types/stateValue.go
askolesov: I would say we have to encode it the same way as it's done in Tendermint/block explorer (cause it's a byte array, it has to be encoded). @Toktar If I'm not mistaken there was a ticket for it.

ankurdotb (dismissed):

go.mod
```diff
@@ -4,7 +4,7 @@ go 1.17
 
 require (
 	github.com/btcsuite/btcutil v1.0.3-0.20201208143702-a53e38424cce
-	github.com/cosmos/cosmos-sdk v0.44.5
+	github.com/cosmos/cosmos-sdk v0.45.1
```
ankurdotb, line 7: Is this a breaking change or not?

ankurdotb (commented):
@askolesov Can I confirm that this is still WIP and not ready for review?

x/cheqd/types/did_validation_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"testing"
+)
+
+func TestIsDid(t *testing.T) {
+	cases := []struct {
+		valid bool
+		did   string
+	}{
+		{true, "did:cheqd:test:wyywywywyw"},
+		{true, "did:cheqd:test:wyywywywyw:sdadasda"},
```
ankurdotb, line 14: I'm unsure of how the rest of the validation works (haven't read ahead), but both of these values are incorrect

Toktar: Yes. PR in draft, Andrew is updating tests now.

x/cheqd/types/validate.go
```diff
@@ -1,72 +1,101 @@
...
+type DIDUrlRule struct {
+	method string
+	allowedNamespaces []string
+}
+
+func NewDIDUrlRule(method string, allowedNamespaces []string) *DIDUrlRule {
+	return &DIDUrlRule{method: method, allowedNamespaces: allowedNamespaces}
+}
+
+func (D DIDUrlRule) Validate(value interface{}) error {
```
ankurdotb: Seems like a typo here

ankurdotb (commented):
I didn't have the chance to fully work through the logic here on allowed DID namespace, but basically it _must_ match the one in the genesis file.

ankurdotb (changes requested):
@askolesov @Toktar @anikitinDSR Massive PR and there is _clearly_ a lot of effort that has gone into this. Thank you all so much!

I've left behind a couple of review comments. (Can do on a call if it's faster.) Some of them might not even need changes, but just an answer/discussion.
 
While this is not in the list of changed files, can we please change the path in `query.proto` L14 from `option (google.api.http).get = "/cheqd/cheqdnode/cheqd/did/{id}";` to `option (google.api.http).get = "/cheqd/v1/did/{id}";` This is more consistent with the Cosmos SDK REST API naming convention.

Also, at what point would/should that be changed from `cheqd/v1` to a different version? Should it be increased in step with module Consensus version, or not?

go.mod
```diff
@@ -3,13 +3,16 @@ module github.com/cheqd/cheqd-node
 go 1.17
 
 require (
+	filippo.io/edwards25519 v1.0.0-beta.2
```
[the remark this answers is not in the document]
anikitinDSR: It's needed just for checking ED25519 key, just for confirming, that publickey represents correct point.

x/cheqd/client/cli/tx_create_did.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/cheqd/cheqd-node/x/cheqd/types"
+	"github.com/cosmos/cosmos-sdk/client"
+	"github.com/cosmos/cosmos-sdk/client/flags"
+	"github.com/cosmos/cosmos-sdk/client/tx"
+	"github.com/spf13/cobra"
+)
+
+func CmdCreateDid() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "create-did [payload-json] [ver-method-id-1] [priv-key-1] [ver-method-id-N] [priv-key-N] ...",
```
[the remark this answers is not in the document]
Toktar: The prefix is added in another place. So, in my terminal it looks good:
```
toktar@toktar-VirtualBox:~/data/cheqd-node/docker/localnet$ cheqd-noded tx cheqd create-did --help
Creates a new DID. [payload-json] is JSON encoded MsgCreateDidPayload. [ver-method-id-N] is the DID fragment that points to the public part of the key in the ledger for the signature N.[priv-key-1] is base base64 encoded ed25519 private key for signature N.If 'interactive' value is used for a key, the key will be read interactively. Prefer interactive mode, use inline mode only for tests.

Usage:
  cheqd-noded tx cheqd create-did [payload-json] [ver-method-id-1] [priv-key-1] [ver-method-id-N] [priv-key-N] ... [flags]

```

go.mod
askolesov: It is a breaking change. Which is not an issue in my opinion because changes in codebase are breaking changes too and require an upgrade proposal anyway.

> Cosmos SDK v0.45.0 is a logical continuation of the v0.44.* series, but brings a couple of state- and API-breaking changes requested by the community.

x/cheqd/keeper/keeper_did.go
```diff
@@ -30,59 +30,64 @@ func (k Keeper) GetDidCount(ctx sdk.Context) uint64 {
...
 	bz := []byte(strconv.FormatUint(count, 10))
 	store.Set(byteKey, bz)
 }
 
 // AppendDid appends a did in the store with a new id and updates the count
-func (k Keeper) AppendDid(ctx sdk.Context, did types.Did, metadata *types.Metadata) (*string, error) {
+func (k Keeper) AppendDid(ctx *sdk.Context, did *types.Did, metadata *types.Metadata) error {
+	// Check that did doesn't exist
+	if k.HasDid(ctx, did.Id) {
+		return types.ErrDidDocNotFound.Wrapf(did.Id)
```
Toktar, line 44: @askolesov Should we change error type here?

askolesov:
> While this is not in the list of changed files, can we please change the path in query.proto L14 from option (google.api.http).get = "/cheqd/cheqdnode/cheqd/did/{id}"; to option (google.api.http).get = "/cheqd/v1/did/{id}"; This is more consistent with the Cosmos SDK REST API naming convention.

Good point, changed

> Also, at what point would/should that be changed from cheqd/v1 to a different version? Should it be increased in step with module Consensus version, or not?

It's up to us. I think we should change the version once we change models. + Ideally, we should keep previous versions for some time.

go.mod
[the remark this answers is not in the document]
askolesov: So we had two options here:
* Try signing and catch panic;
* Use `filippo.io/edwards25519` <- we choose this option.

Do you think there is a better way to do it?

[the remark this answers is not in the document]
askolesov: One more thing: if there is a bug in the library (in which I doubt because it's an extended copy of the internal one) and an invalid key passes static validation, signatures will be checked using standard `ed25519` library and the request will fail.

ankurdotb (approved):

tests/e2e-complex/upgrade/common.sh
```diff
@@ -23,20 +23,19 @@ AMOUNT_BEFORE="19000000000000000"
...
-    -e HOME=/cheqd \
-    --entrypoint "cheqd-noded" \
-    ${CHEQD_IMAGE_FROM} "$@"
+    docker run --rm \
+        -v "$(pwd):/cheqd" \
+        --network host \
+        -u root \
+        -e HOME=/cheqd \
+        --entrypoint "cheqd-noded" \
+        ${CHEQD_IMAGE_FROM} "$@"
```
ankurdotb, line 38: Shouldn't we use a similar Docker compose up technique like the pipeline? + not a good idea to test as root (cheqd user should be default in the build process, so we could change this to cheqd)

tests/e2e-complex/upgrade/upgrade_and_check.sh
```diff
@@ -5,6 +5,16 @@ set -euox pipefail
...
+cheqd_noded_docker() {
+    docker run --rm \
+        -v "$(pwd):/home/cheqd" \
+        --network host \
+        -u root \
+        -e HOME=/home/cheqd \
+        --entrypoint "cheqd-noded" \
+        ${CHEQD_IMAGE_TO} "$@"
+}
+
```
ankurdotb, line 17: Same as the other Docker comment

ankurdotb (changes requested):

tests/e2e-complex/upgrade/common.sh
askolesov: It's the CLI image. I suggest changing this to direct cheqd-cli calls in future. cc @anikitinDSR. And yes, probably we could use the latest docker-compose as well.

askolesov: I'm going to create a follow-up ticket and merge in order to make a release and start testing. This test is isolated so we can improve it later.

askolesov: Ticket: https://app.clickup.com/t/28t9wfg

ankurdotb (dismissed):

Toktar:
Approved by @Toktar

PR #286, opened 2022-02-21
DEV-970: Allow only base58 16 or 32 symbols DIDs

anikitinDSR:


ankurdotb:
Task linked: [DEV-970 Fix bugs in validation of did:cheqd unique identifiers format](https://app.clickup.com/t/6600954/DEV-970)

tests/e2e/helpers.py
```diff
@@ -12,7 +12,7 @@
 ENCODING = "utf-8"
 READ_BUFFER = 60000
 
-TEST_NET_NETWORK = "cheqd-testnet-2"
+TEST_NET_NETWORK = "cheqd-testnet-4"
 LOCAL_NET_NETWORK = "cheqd"
 TEST_NET_NODE_TCP = "--node 'tcp://seed1.us.testnet.cheqd.network:26657'"
 TEST_NET_NODE_HTTP = "--node http://node1.eu.testnet.cheqd.network:26657/"
```
ankurdotb, line 18: Change to `tcp://rpc.testnet.cheqd.network`and https://rpc.testnet.cheqd.network`. We no longer expose port 26657, and it's behind HTTPS on port 443.

tests/e2e/test_identity.py
```diff
@@ -118,7 +118,7 @@ def test_did_query_non_existent():
 
 
 def test_did_wrong_version_update():
-    did = fr"did:cheqd:testnet:{random_string(5)}"
+    did = fr"did:cheqd:testnet:{random_string(16)}"
```
ankurdotb, line 121: See definition in the `did:indy` link I provided.

x/cheqd/tests/constants.go
```diff
@@ -1,16 +1,18 @@
...
-	BobKey1     = BobDID + "#key-1"
-	BobKey2     = BobDID + "#key-2"
-	BobKey3     = BobDID + "#key-3"
-	BobKey4     = BobDID + "#key-4"
-	CharlieKey1 = CharlieDID + "#key-1"
-	CharlieKey2 = CharlieDID + "#key-2"
-	CharlieKey3 = CharlieDID + "#key-3"
+	AliceDID      = "did:cheqd:test:00000000000alice"
+	BobDID        = "did:cheqd:test:0000000000000bob"
+	CharlieDID    = "did:cheqd:test:000000000charlie"
```
ankurdotb, line 6: Should this be testnet (to exercise namespace check)? Basically, using `test` as the namespace is fine, as long as that's the same namespace used in the test. For example, if namespace is set to "mainnet", a user shouldn't be able to create a DID with namespace "testnet" or "test" and so on. (I don't know whether we check this already, if not, the probably needs to be its own task next sprint.)

x/cheqd/utils/did.go
```diff
@@ -6,6 +6,7 @@ import (
 )
 
 var DidForbiddenSymbolsRegexp, _ = regexp.Compile(`^[^#?&/\\]+$`)
+var DidAllowedSymbolsRegexp, _ = regexp.Compile(`^[a-zA-Z0-9_]*$`)
```
ankurdotb, line 9: I would use an actual Base58 check, instead of relying on just checking the character set: https://github.com/anaskhan96/base58check (there might be other libraries). For example, Base58 _doesn't_ allow ambiguous characters, such as `0` vs 'O` etc. https://en.wikipedia.org/wiki/Binary-to-text_encoding#Base58

See Base58 alphabet here: https://en.bitcoinwiki.org/wiki/Base58#Alphabet_Base58

I don't think you need to check for forbidden symbols, just check for allowed symbols.

x/cheqd/utils/did_test.go
```diff
@@ -10,8 +10,8 @@ func TestIsDid(t *testing.T) {
 		valid bool
 		did   string
 	}{
-		{true, "did:cheqd:test:wyywywywyw"},
-		{true, "did:cheqd:test:wyywywywyw:sdadasda"},
+		{true,  "did:cheqd:test:000000wyywywywyw"},
```
ankurdotb, line 13: This is a great example of something that includes no special characters, but fails Base58 encoding spec.

ankurdotb (changes requested):

ankurdotb (approved):

ankurdotb (approved):

anikitinDSR:
Let's pause it by PR:  #285

askolesov:
Superseded by https://github.com/cheqd/cheqd-node/pull/285

PR #303, opened 2022-03-24
DEV-520 - Chore: Added Github CodeQL Action

jay-dee7:
This PR adds a new Github Action - CodeQL.

This Github Action does static analysis of our code and reports for any gotchas like integer overflow, security vulnerabilities, bugs, etc


ankurdotb:
Task linked: [DEV-520 Set up Github CodeQL Analysis for cheqd-node repo](https://app.clickup.com/t/6600954/DEV-520)

jay-dee7:
guys, CodeQL seems to working as expected, do we want to do something more or this is enough?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,41 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
```
ankurdotb, line 7: Maybe also ditch this?

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,41 @@
...
+      with:
+        languages: ${{ matrix.language }}
+        # this covers a larger portion of checks but are experimental right now
+        queries: security-and-quality
+    - run: |
+       make build
+
+    - name: Perform CodeQL Analysis
+      uses: github/codeql-action/analyze@v1
+
```
ankurdotb, line 41: Not sure if this will work - does it work on compiled code? The make/run instructions are different for our code as far as I remember.

ankurdotb (changes requested):
Instead of having it in a separate workflow, I would recommend adding this to either `test.yml` (or maaaaaaaaybe `lint.yml`). This is because in terms of chaining multiple GitHub Actions (which I want to do), the maximum depth of workflow chains that can be called is 3 workflows. I see this as going lint.yml -> test.yml (which is actually test and build) -> release.yml

Maybe sync up with @anikitinDSR who is working on new release pipelines and just generally pipelines.

.github/workflows/codeql.yml
jay-dee7: shouldn't we keep this? it would run when we make new PRs or should we keep it like
```yaml
on:
  pull_request:
```

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,41 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ main ]
+  pull_request:
+    branches: [ main ]
+  schedule:
+    - cron: '15 3 * * 4'
```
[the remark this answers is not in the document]
jay-dee7: CodeQL Team does recommend this running on a schedule just to track vulnerabilities on a regular basis but maybe we can remove it since out PR rate is high?

.github/workflows/codeql.yml
jay-dee7: it does work on program files and compiled binaries. More about this here: https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-the-codeql-workflow-for-compiled-languages#about-autobuild-for-codeql

jay-dee7: and I did check it with Tasos and also cross  checked the Makefile, `make build` is how we code, right?

jay-dee7:
TIL: CodeQL makes your Code a database and then queries your code like data

.github/workflows/lint.yml
```diff
@@ -3,6 +3,31 @@ name: Lint
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      with:
+        languages: ${{ matrix.language }}
+        # this covers a larger portion of checks but are experimental right now
+        queries: security-and-quality
+    - run: make build
```
askolesov, line 28: Why do we need to build here?

jay-dee7: @askolesov forgot to update this PR, I thought I already did. added codeql to the test.yml file

askolesov: I mean can we run codeql in a separate job that doesn't build our sources?

.github/workflows/codeql.yml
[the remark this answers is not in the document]
askolesov: I think PR rate is high enough

.github/workflows/codeql.yml
askolesov: Maybe we used it wrong, but `pull_request` didn't trigger on pushes

.github/workflows/codeql.yml
askolesov: Still not sure that we should build manually

askolesov (dismissed):
IMO the CodeQL job should be in a separate workflow. It doesn't look good neither in `lint` because it builds code nor in `test` because it is not about testing. Trigger can be push instead of cron.

@ankurdotb @jay-dee7 What are your thoughts?

jay-dee7:
> 

I was thinking about keeping this in separate workflow but @ankurdotb suggested that we wanted to chain workflows together and @anikitinDSR suggested to keep it in test workflow since we already do a build there. But now that I think of it, it does not belong in test workflow 😕

.github/workflows/lint.yml
jay-dee7: we can do it without building but we'll just be missing a check. CodeQL can analyse build process and the output binary so we should definitely use it

.github/workflows/codeql.yml
jay-dee7: @ankurdotb @anikitinDSR can you guys share what you think?

.github/workflows/codeql.yml
jay-dee7: oh really? let me check that

ankurdotb: @askolesov BTW this is incorrect. `pull_request` **also** triggers on every push to the branch associated with the pull request

ankurdotb (approved):

PR #318, opened 2022-05-03
gofumpt -w -l .

faddat:
excellent code formatting for golang

go install mvdan.cc/gofumpt@latest
gofumpt -w -l .


ankurdotb:
@askolesov Seems like a good chore, can you take a look please and see if you agree?

askolesov:
Yeah, looks good. I suggest then adding a linter rule to check formatting. @ankurdotb Would you agree?

ankurdotb:
> Yeah, looks good. I suggest then adding a linter rule to check formatting. @ankurdotb Would you agree?

Yep sounds good to me! Maybe as a step in the Go lint job?

askolesov:
@faddat Thanks! Wasn't able to commit to your branch, so created another PR: https://github.com/cheqd/cheqd-node/pull/324. Your commits are preserved. In future, please allow pushes - there should be a checkbox when you create a PR.

PR #322, opened 2022-05-16
Fix workflow calls to refer local versions instead of master

askolesov:


rosspower11 (approved):

askolesov:
@ankurdotb You said it was doe intentionally. What is the point? If workflows are not changed in a PR, it doesn't matter what version to run. If they are, isn't it better to make sure that the new version works?

PR #336, opened 2022-07-01
test: Add resources to upgrade integration tests

Toktar:


tests/e2e-complex/upgrade/common.sh
```diff
@@ -159,6 +161,38 @@ function send_did_new () {
...
+    RESOURCE_DATA='{ "content": "test data" }';
+
+    # Post the message
+    # shellcheck disable=SC2086
+    resource=$(local_client_tx tx resource create-resource \
+    --collection-id ${collection_id_to_write} \
+    --resource-id ${resource_to_write} \
+    --resource-name "${RESOURCE_NAME}" \
+    --resource-type "${RESOURCE_RESOURCE_TYPE}" \
+    --resource-file (echo "${RESOURCE_DATA}") \
```
ankurdotb, line 188: Can you echo inline like this or should it be path to a file?

Toktar: No, `resource-file` flag can use only files

PR #337, opened 2022-07-01
fix: Bump Cosmos SDK to v0.45.6

Toktar:


ankurdotb:
Not bumping to v0.45.6

PR #341, opened 2022-07-05
fix: Get rid of debian package for release

anikitinDSR:


askolesov:
@anikitinDSR Should we remove deb-build script as well?

askolesov (approved):

PR #373, opened 2022-08-17
test: Move unit testing to Gingko [DEV-1652]

anikitinDSR:


drgomesp:
Nice work, I like Gingko as well! BTW, did you guys know or had a chance to look at [GoConvey](https://github.com/smartystreets/goconvey)? It's a very interesting alternative for more behavioral-oriented testing. Could be interesting to have a look and see if it could potentially help us in the direction we are aiming for. 

Well done!

x/cheqd/tests/setup/config.go
```diff
@@ -0,0 +1,3 @@
+package setup
+
+const DID_NAMESPACE = "test"
```
ankurdotb, line 3: Can set to testnet?

x/cheqd/tests/tests_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestTestsGeneral(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Cheqd Module")
```
ankurdotb, line 12: cheqd DID module

x/cheqd/types/did_verification_method_test.go
```diff
@@ -8,290 +8,303 @@ import (
...
+			Expect(err).To(BeNil())
+		} else {
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring(errorMsg))
+		}
+	})
+	When("verification method with expected multibase key", func() {
+		It("is valid", func() {
+			struct_ = VerificationMethod{
 				Id:                 "did:cheqd:aaaaaaaaaaaaaaaa#qwe",
```
ankurdotb, line 47: - Doesn't contain namespace
- Use a realistic DID. Check #399 and see the discussion there but it should be 22.

x/cheqd/types/stateValue_test.go
```diff
@@ -1,81 +1,90 @@
...
-		Deactivated: false,
-		VersionId:   utils.GetTxHash(ctx2.TxBytes()),
-	}
-
-	metadata := NewMetadataFromContext(ctx1)
-	metadata.Update(ctx2)
-
-	require.Equal(t, expectedMetadata, metadata)
-}
+var _ = Describe(`StaeValue tests`, func() {
```
ankurdotb, line 18: `StateValue tests` <- there's a typo in above

x/cheqd/types/tx_msg_update_did_test.go
```diff
@@ -50,13 +68,15 @@ func TestMsgUpdateDidValidation(t *testing.T) {
...
-		{
-			name: "negative: version id is required",
-			struct_: &MsgUpdateDid{
+			}
+			isValid = false
+			errorMsg = "payload: (authentication: there should be no duplicates.).: basic validation failed"
+		})
+	})
+
+	When("VersionId is empty", func() {
```
ankurdotb, line 77: Isn't VersionID set by ledger?

x/cheqd/utils/did_url_test.go
```diff
@@ -1,93 +1,80 @@
...
 )
 
-func TestIsDidURL(t *testing.T) {
-	cases := []struct {
-		name   string
-		valid  bool
-		didUrl string
-	}{
+var _ = Describe("DID-URL tests", func() {
+	DescribeTable("Check the DID-URL join functionality (without functionality)",
```
ankurdotb, line 10: This whole section is incorrect

ankurdotb (changes requested):
There is a LOT to fix here. Some of the identifier tests, for example, could be done in the scope of #399. However, there's a more basic issue here first: the tests in the pipeline doesn't run.

x/cheqd/utils/did_url_test.go
askolesov: Can be follow-up ticket

x/resource/keeper/msg_server_create_resource.go
```diff
@@ -46,8 +46,8 @@ func (k msgServer) CreateResource(goCtx context.Context, msg *types.MsgCreateRes
 
 	// Build Resource
 	resource := msg.Payload.ToResource()
-
-	resource.Header.Checksum = sha256.New().Sum(resource.Data)
+	checksum := sha256.Sum256([]byte(resource.Data))
+	resource.Header.Checksum = checksum[:]
```
askolesov, line 50: What is this for?

.github/workflows/dispatch.yml
```diff
@@ -9,23 +9,22 @@ jobs:
...
-  # call-test:
-  #   name: "Test"
-  #   needs: call-build
-  #   uses: ./.github/workflows/test.yml
-  #   secrets: inherit
+  call-test:
+    name: "Test"
+    needs: call-build
+    uses: ./.github/workflows/test.yml
+    secrets: inherit
```
askolesov, line 23: Should we inherit secrets in all steps?

.github/workflows/test.yml
```diff
@@ -25,9 +25,12 @@ jobs:
 
       - name: Install ginkgo
         run: |
-          go get github.com/onsi/ginkgo/v2/ginkgo
+          go install github.com/onsi/ginkgo/v2/ginkgo
```
askolesov, line 28: Please so is the same way as in integration tests PR.

askolesov (changes requested):

x/cheqd/types/did_verification_method_test.go
askolesov: * Currently, namespace is mandatory. We can make it optional and restore during requests normalization. Better to do in a separate ticket.
* `Use a realistic DID` Imo tests should be fixed in the corresponding PR (the one that fixes identifier validation).

x/cheqd/types/did_verification_method_test.go
```diff
@@ -8,290 +8,303 @@ import (
...
+
+	BeforeEach(func() {
+		struct_ = VerificationMethod{}
+		baseDid = ""
+		allowedNamespaces = []string{}
+		isValid = false
+		errorMsg = ""
+	})
+
+	AfterEach(func() {
```
askolesov: Le'ts use tables instead

x/cheqd/types/tx_msg_update_did_test.go
anikitinDSR: We have to set VersionID for avoiding race condition and for point which particular "snapshot" of DID we want to edit/update

askolesov (approved):

PR #374, opened 2022-08-17
fix(installer): Change snapshot URL calculating logic

anikitinDSR:

### DIFF
--- a/installer/installer.py
+++ b/installer/installer.py
@@ -64,8 +64,8 @@ SEEDS_FILE = "https://raw.githubusercontent.com/cheqd/cheqd-node/main/networks/{
 ###############################################################
 DEFAULT_SNAPSHOT_SERVER = "https://snapshots-cdn.cheqd.net"
 DEFAULT_INIT_FROM_SNAPSHOT = "yes"
-TESTNET_SNAPSHOT = "https://snapshots-cdn.cheqd.net/testnet/{}/cheqd-testnet-4_{}.tar.gz"
-MAINNET_SNAPSHOT = "https://snapshots-cdn.cheqd.net/mainnet/{}/cheqd-mainnet-1_{}.tar.gz"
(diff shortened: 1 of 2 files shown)


installer/installer.py
```diff
@@ -64,9 +64,10 @@
 ###############################################################
 DEFAULT_SNAPSHOT_SERVER = "https://snapshots.cheqd.net"
 DEFAULT_INIT_FROM_SNAPSHOT = "yes"
-TESTNET_SNAPSHOT = "https://cheqd-node-backups.ams3.cdn.digitaloceanspaces.com/testnet/latest/cheqd-testnet-4_{}.tar.gz"
-MAINNET_SNAPSHOT = "https://cheqd-node-backups.ams3.cdn.digitaloceanspaces.com/mainnet/latest/cheqd-mainnet-1_{}.tar.gz"
-CHECKSUM_URL_BASE = "https://cheqd-node-backups.ams3.cdn.digitaloceanspaces.com/"
+TESTNET_SNAPSHOT = "https://snapshots-cdn.cheqd.net/testnet/{}/cheqd-testnet-4_{}.tar.gz"
+MAINNET_SNAPSHOT = "https://snapshots-cdn.cheqd.net/mainnet/{}/cheqd-mainnet-1_{}.tar.gz"
+MAX_SNAPSHOT_DAYS = 7
+CHECKSUM_URL_BASE = "https://snapshots-cdn.cheqd.net/"
```
ankurdotb, line 70: This is called in line 515. Our current logic always assumes the checksum file is in the `latest` folder, but that's no longer the case. 

This is line 515:
`checksum_url = os.path.join(CHECKSUM_URL_BASE, self.interviewer.chain, "latest/md5sum.txt")`

So obviously we didn't download the full snapshot, but it would have failed later at the checksum stage for sure.

ankurdotb (changes requested):

ankurdotb (approved):

PR #396, opened 2022-09-28
fix: Checksum calculation [DEV-1649]

anikitinDSR:


ankurdotb:
Task linked: [DEV-1649 Fix checksum calculation on the ledger side](https://app.clickup.com/t/6600954/DEV-1649)

app/migration_v_0_7.go
```diff
@@ -0,0 +1,20 @@
+package app
```
ankurdotb, line 1: Same comment on migration handler

x/resource/keeper/msg_server_create_resource.go
```diff
@@ -47,7 +47,9 @@ func (k msgServer) CreateResource(goCtx context.Context, msg *types.MsgCreateRes
 	// Build Resource
 	resource := msg.Payload.ToResource()
 
-	resource.Header.Checksum = sha256.New().Sum(resource.Data)
+	h := sha256.New()
+	h.Write(resource.Data)
+	resource.Header.Checksum = h.Sum(nil)
```
ankurdotb, line 52: Not sure if this logic is correct. Isn't the correct function for this https://pkg.go.dev/crypto/sha256#New?

ankurdotb (changes requested):

x/resource/keeper/msg_server_create_resource.go
Eengineer1: Hey, yeah, that's the correct one.

```go
sha256.Sum256(resource.Data)
```

This outputs a fixed length array type of [32]byte.
So the problem lies in that Protobuf definitions don't allow for fixed byte array lengths, since they are primitively encoded.

I'd suggest creating an addressable (in-memory) array and converting it to a slice that points to those 32 byte values, as below:

```go
checksum := sha256.Sum256([]byte(resource.Data))
resource.Header.Checksum = checksum[:]
```

Eengineer1 (commented):
Hey, yeah, that's the correct one.

```go
sha256.Sum256(resource.Data)
```

This outputs a fixed length array type of [32]byte.
So the problem lies in that Protobuf definitions don't allow for fixed byte array lengths, since they are primitively encoded.

I'd suggest creating an addressable (in-memory) array and converting it to a slice that points to those 32 byte values, as below:

```go
checksum := sha256.Sum256([]byte(resource.Data))
resource.Header.Checksum = checksum[:]
```

Came across this while switching test suites to Gingko, so it's fixed there as well.

x/resource/keeper/msg_server_create_resource.go
anikitinDSR: Hey! Both of them are the same and do the same steps inside but I agree It would be more readable and understandable to use just `sha256.Sum256()`

app/migration_v_0_7.go
```diff
@@ -0,0 +1,19 @@
...
+
+import (
+	"crypto/sha256"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func (app *App) Migration07(ctx sdk.Context) {
+	all_resources := app.resourceKeeper.GetAllResources(&ctx)
```
askolesov, line 10: Loading everything into memory is not the best approach. Resources can be large. I would suggest to use iterator instead and load resources one by one.

x/resource/tests/setup/setup.go
```diff
@@ -89,3 +90,8 @@ func Setup() TestSetup {
 	setup.Keeper.SetDidNamespace(&ctx, cheqdsetup.DID_NAMESPACE)
 	return setup
 }
+
+func CreateChecksum(data []byte) []byte {
```
askolesov, line 94: I would either inline this or rename to highlight that it's sha256 checksumm.

askolesov (changes requested):

app/migration_v_0_7.go
askolesov: Agreed to do it in a separate PR dedicated to migrations

app/migration_v_0_7.go
askolesov: Left a todo in code

askolesov (approved):

askolesov (approved):

PR #397, opened 2022-09-30
fix: Add resourceType for versioning functionality [DEV-1703]

anikitinDSR:
Signed-off-by: Andrew Nikitin <andrew.nikitin@cheqd.io>


ankurdotb:
Task linked: [DEV-1703 Implement Resource versioning fix to include Resourcetype as part of linking of versions](https://app.clickup.com/t/6600954/DEV-1703)

app/migration_v_0_7.go
```diff
@@ -0,0 +1,24 @@
+package app
```
ankurdotb, line 1: FYI, our next package version will be 1.x.x since it's breaking from 0.6.x and we ended up triggering a 0.7.x release. This will also bring it in line with semantic versioning, so this handler will need to be renamed.

ankurdotb (changes requested):

app/migration_v_0_7.go
```diff
@@ -0,0 +1,24 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// Here we need to make all the links the
+func (app *App) Migration07(ctx sdk.Context) {
+	resourceList := app.resourceKeeper.GetAllResources(&ctx)
```
askolesov, line 9: We need to:
* Firstly, iterate without loading everything in memory and completely erase previous/next links;
* Next, iterate without loading everything in memory and call  `GetLastResourceVersionHeader`;

askolesov (changes requested):

app/migration_v_0_7.go
askolesov: Will fix it later in a separate PR.

app/migration_v_0_7.go
askolesov: Ok, will do it later.

askolesov (approved):

PR #398, opened 2022-10-02
feat(proto): Proto API v2 [DEV-1786]

askolesov:


proto/cheqd/did/v2/diddoc.proto
```diff
@@ -0,0 +1,44 @@
...
+  string id = 1;
+  string type = 2;
+  string controller = 3;
+  string verification_material = 4;
+}
+
+message Service {
+  string id = 1;
+  string type = 2;
+  string service_endpoint = 3;
```
ankurdotb, line 31: This can either be a string, map, or set, so is this better as repeated? https://w3c.github.io/did-core/#services

proto/cheqd/did/v2/diddoc.proto
```diff
@@ -0,0 +1,44 @@
...
+message DidDocWithMetadata {
+  DidDoc did_doc = 1;
+  Metadata metadata = 2;
+}
+
+message Metadata {
+  string created = 1;
+  string updated = 2;
+  bool deactivated = 3;
+  string version_id = 4;
```
ankurdotb, line 43: Would it be possible to have a nextVersionId? https://w3c.github.io/did-core/#did-document-metadata

Also while we are thinking of protobufs, how can we allow for querying historical DIDDoc states? Perhaps we add previousVersionId and nextVersionId and start using UUIDs instead of tx hash? What do you think?

proto/cheqd/did/v2/query.proto
```diff
@@ -0,0 +1,22 @@
...
+package cheqd.did.v2;
+
+import "google/api/annotations.proto";
+import "cheqd/did/v2/diddoc.proto";
+
+option go_package = "github.com/cheqd/cheqd-node/x/cheqd/types/v2";
+
+service Query {
+	rpc DidDoc(QueryGetDidDocRequest) returns (QueryGetDidDocResponse) {
+		option (google.api.http).get = "/cheqd/v1/diddoc/{id}";
```
ankurdotb, line 12: Seems incorrect. Should be `cheqd/did/v2/diddoc` right?

proto/cheqd/resource/v2/query.proto
```diff
@@ -0,0 +1,58 @@
...
+  rpc ResourceMetadata(QueryGetResourceMetadataRequest) returns (QueryGetResourceMetadataResponse) {
+    option (google.api.http).get = "/1.0/identifiers/{collection_id}/resources/{id}/metadata";
+  }
+  rpc CollectionResources(QueryGetCollectionResourcesRequest) returns (QueryGetCollectionResourcesResponse) {
+    option (google.api.http).get = "/1.0/identifiers/{collection_id}/resources/all";
+  }
+  rpc AllResourceVersions(QueryGetAllResourceVersionsRequest) returns (QueryGetAllResourceVersionsResponse) {
+    option (google.api.http).get = "/1.0/identifiers/{collection_id}/resources/versions/{name}";
+  }
+}
```
ankurdotb, line 23: So I would simplify this. You don't need to handle all of these as a ledger API, this is the API path that our DID Resolver needs to support. From the ledger, just have a fetch for resource and fetch collection metadata endpoints.

So I'd say:

- `cheqd/resource/v2/resource/<resource-id>` fetch resource
- `cheqd/resource/v2/metadata/<collection-id-OR-resource-id>` fetch collection/resource metadata

What do you think?

ankurdotb (dismissed):

ankurdotb:
Task linked: [DEV-1786 Build and lint Protobuf definitions for v2](https://app.clickup.com/t/6600954/DEV-1786)

proto/cheqd/did/v1/did.proto
```diff
@@ -4,7 +4,7 @@ package cheqdid.cheqdnode.cheqd.v1;
 option go_package = "github.com/cheqd/cheqd-node/x/cheqd/types";
 
 import "google/protobuf/any.proto";
-import "cheqd/v1/common.proto";
+import "cheqd/did/v1/common.proto";
```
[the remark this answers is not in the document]
askolesov: This is a fully backward compatible change. No need to update clients. File structure looks awful otherwise.

proto/cheqd/did/v2/diddoc.proto
askolesov: Good catch

proto/cheqd/did/v2/diddoc.proto
askolesov: Can only say that this is a highly complex topic that requires a separate research. I don't see any easy ways of querying historical data.

I've just realized that using tx hash is wrong because several updates can happen within a single transaction.

proto/cheqd/did/v2/query.proto
askolesov: Yeah... I left paths for future because they are not breaking changes either (+ nobody uses HTTP gateway).

proto/cheqd/resource/v2/query.proto
askolesov: > cheqd/resource/v2/resource/<resource-id>

This is hard because we have index by `collection id` and only then by `resource id`. There is no global index by `resource id`.

> cheqd/resource/v2/metadata/<collection-id-OR-resource-id>

Same reason here + there is no way to find out which ID is provided - of a collection or of a resource.

proto/cheqd/resource/v2/query.proto
```diff
@@ -0,0 +1,58 @@
...
+    option (google.api.http).get = "/cheqd/resource/v2/collection/{collection_id}/resources/{id}";
+  }
+  rpc ResourceMetadata(QueryGetResourceMetadataRequest) returns (QueryGetResourceMetadataResponse) {
+    option (google.api.http).get = "/cheqd/resource/v2/collection/{collection_id}/resources/{id}/metadata";
+  }
+  rpc CollectionResources(QueryGetCollectionResourcesRequest) returns (QueryGetCollectionResourcesResponse) {
+    option (google.api.http).get = "/cheqd/resource/v2/collection/{collection_id}/resources";
+  }
+  rpc AllResourceVersions(QueryGetAllResourceVersionsRequest) returns (QueryGetAllResourceVersionsResponse) {
+    option (google.api.http).get = "/cheqd/resource/v2/collection/{collection_id}/resources/versions/{name}";
```
ankurdotb, line 21: Remove this

ankurdotb (changes requested):

proto/cheqd/resource/v2/fee.proto
```diff
@@ -0,0 +1,22 @@
...
+import "gogoproto/gogo.proto";
+import "cosmos/base/v1beta1/coin.proto";
+import "cosmos_proto/cosmos.proto";
+
+// FeeParams defines the parameters for the `resource` module fixed fee.
+message FeeParams {
+    // Media types define the fixed fee each for the `resource` module.
+    cosmos.base.v1beta1.Coin image = 1 [(gogoproto.nullable) = false];
+    cosmos.base.v1beta1.Coin json = 2 [(gogoproto.nullable) = false];
+    cosmos.base.v1beta1.Coin default = 3 [(gogoproto.nullable) = false];
```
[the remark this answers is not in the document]
Eengineer1: Switched to typed map that allows `string` -> `sdk.Coin` definitions.

Eengineer1 (commented):
Added revisited `FeeParams` definitions for interchangeability through gov proposals.

We need to mind for `tx_types` to be snake cased or defined per MsgTypeUrl.
As for resources, `media_types` keys can easily be valid mime types.

askolesov:
Processed the rest of comments. I think it's ready now.

askolesov:
Todo: Add deactivation transaction

PR #401, opened 2022-10-07
test: Move integration tests to ginkgo [DEV-1784]

askolesov:


ankurdotb:
Task linked: [DEV-1784 Implement new approach for integration (system) tests](https://app.clickup.com/t/6600954/DEV-1784)

.github/workflows/build.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         uses: actions/upload-artifact@v3
         with:
           name: cheqd-noded
-          path: dist/**/cheqd-noded
+          path: dist/ubuntu-latest-amd64_linux_amd64_v1/cheqd-noded
```
anikitinDSR, line 36: Why did you setup the exactly path instead of just `**`?

tests/integration/cli_did_negative_test.go
```diff
@@ -0,0 +1,406 @@
...
+		// Fail to create a new DID Doc with non-supported VM type
+		payload3 := payload2
+		payload3.VerificationMethod[0].Type = "NonSupportedVMType"
+		_, err = cli.CreateDid(payload3, signInputs2, testdata.BASE_ACCOUNT_2)
+		Expect(err).ToNot(BeNil())
+
+		// Fail to create a new DID Doc with already existing DID
+		_, err = cli.CreateDid(payload, signInputs, testdata.BASE_ACCOUNT_1)
+		Expect(err).ToNot(BeNil())
+	})
```
ankurdotb, line 144: This whole section seems like different scenarios/use cases, and also different than the one on top. Describing it as a comment is not helping distinguish differences in these test cases, as far as I can see. What am I missing here?

tests/integration/cli_did_negative_test.go
```diff
@@ -0,0 +1,406 @@
...
+			{
+				VerificationMethodId: keyId,
+				PrivKey:              privKeyFuzzedExtra,
+			},
+			{
+				VerificationMethodId: keyId2AsExtraController,
+				PrivKey:              privKey2,
+			},
+		}, testdata.BASE_ACCOUNT_1)
+		Expect(err).ToNot(BeNil())
```
ankurdotb, line 362: There's a lot of repetition here to provide the different inputs required, which makes the test hard to read. Will this be better off [as a table, defined once, and used across these tests](https://onsi.github.io/ginkgo/#table-specs)?

tests/integration/cli_did_test.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/cheqd/cheqd-node/x/cheqd/types"
+	"github.com/google/uuid"
+	"github.com/multiformats/go-multibase"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("cheqd cli", func() {
+	It("can create diddoc, update it and query the result", func() {
```
ankurdotb, line 21: These are very different actions and this is an overloaded test. Especially once we have pricing differentiation, where we _do_ care what kind of DID transaction it is.

tests/integration/cli_did_test.go
```diff
@@ -0,0 +1,106 @@
...
+		// Update the DID Doc
+		newPubKey, newPrivKey, err := ed25519.GenerateKey(nil)
+		Expect(err).To(BeNil())
+
+		newPubKeyMultibase58, err := multibase.Encode(multibase.Base58BTC, newPubKey)
+		Expect(err).To(BeNil())
+
+		payload2 := types.MsgUpdateDidPayload{
+			Id: did,
+			VerificationMethod: []*types.VerificationMethod{
```
ankurdotb, line 65: I don't see tests done with different kinds of keys, e.g., JsonWebKey2020 vs Ed25519, or trying to use incorrect/unsupported key types (I didn't see an unsupported key type test in negative either)

tests/integration/cli_did_test.go
```diff
@@ -0,0 +1,106 @@
...
+		Expect(res2.Code).To(BeEquivalentTo(0))
+
+		// Query the DID Doc
+		resp, err := cli.QueryDid(did)
+		Expect(err).To(BeNil())
+
+		didDoc := resp.Did
+		Expect(didDoc.Id).To(BeEquivalentTo(did))
+		Expect(didDoc.Authentication).To(HaveLen(1))
+		Expect(didDoc.Authentication[0]).To(BeEquivalentTo(keyId))
```
ankurdotb, line 99: Correct in this case, but not necessarily true?

tests/integration/cli_did_test.go
```diff
@@ -0,0 +1,106 @@
...
+		Expect(didDoc.VerificationMethod).To(HaveLen(1))
+		Expect(didDoc.VerificationMethod[0].Id).To(BeEquivalentTo(keyId))
```
ankurdotb, line 101: Correct in this case, but what about multiple verification methods?

tests/integration/cli_did_test.go
```diff
@@ -0,0 +1,106 @@
...
+		Expect(didDoc.VerificationMethod[0].Type).To(BeEquivalentTo("Ed25519VerificationKey2020"))
```
ankurdotb, line 102: Doesn't account for different key types

tests/integration/cli_did_test.go
```diff
@@ -0,0 +1,106 @@
...
+		Expect(didDoc.VerificationMethod[0].Controller).To(BeEquivalentTo(did))
```
ankurdotb, line 103: True in this case, but this is not always true

tests/integration/cli_resource_negative_test.go
```diff
@@ -0,0 +1,222 @@
...
+		// Fail to query resource collection with missing cli arguments
+		//   a. missing collection id
+		_, err := cli.QueryResourceCollection("")
+		Expect(err).To(HaveOccurred())
+
+		// Fail to query resource collection with non-existing collection id
+		_, err = cli.QueryResourceCollection(collectionId2)
+		Expect(err).To(HaveOccurred())
+	})
+})
```
ankurdotb, line 222: Quite a few different scenarios but hard to understand which ones passed and didn't. (Unless I'm missing some output file somewhere?)

tests/integration/cli_resource_test.go
```diff
@@ -0,0 +1,158 @@
...
+			},
+			Authentication: []string{keyId},
+		}
+
+		signInputs := []cli_types.SignInput{
+			{
+				VerificationMethodId: keyId,
+				PrivKey:              privKey,
+			},
+		}
```
ankurdotb, line 50: Same as above. Would a table be a less verbose way of describing this?

ankurdotb (changes requested):
Unless I'm massively missing something obvious, I don't see a simple-to-understand outcome of what passed vs what failed (maybe this requires exporting to JSON or JUnit). Even before we get to that though, test readability is low because happy/unhappy path doesn't cover a lot of scenarios and/or blends different kinds of scenarios all into the same Ginkgo test execution node.

tests/integration/cli_did_negative_test.go
```diff
@@ -0,0 +1,406 @@
...
+	cli_types "github.com/cheqd/cheqd-node/x/cheqd/client/cli"
+	"github.com/cheqd/cheqd-node/x/cheqd/types"
+	"github.com/google/uuid"
+	"github.com/multiformats/go-multibase"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("cheqd cli negative", func() {
+	It("cannot create diddoc with missing cli arguments, sign inputs mismatch, non-supported VM type, already existing did", func() {
```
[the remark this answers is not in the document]
anikitinDSR: As the first step we could use a JUnit reports for this. As I use for unit tests.

tests/integration/cli_resource_negative_test.go
Eengineer1: There are relationships among different bundled cases.
That's why a procedural / functional style allows for less repetition.

One refactor that allows to block-scope procedural scenarios is to use `Context()` and enclose assertions in different `It()` leaf nodes.

tests/integration/cli_resource_test.go
Eengineer1: Same as above, since container nodes change declaration scope.

Eengineer1 (changes requested):
Keeping this as open for integration tests improvements, as it allows for retaining conversations and code state till we pick them up in the near future, after merging the current state to `develop`.

PR #424, opened 2022-10-25
build: Release with v-tags

ankurdotb:


cmd/cheqd-noded/cmd/utils.go
```diff
@@ -38,7 +38,7 @@ func readCosmConfig(homeDir string) (cosmcfg.Config, error) {
 		return cosmcfg.Config{}, fmt.Errorf("failed to read in app.toml: %w", err)
 	}
 
-	config := cosmcfg.GetConfig(v)
+	config, _ := cosmcfg.GetConfig(v)
```
askolesov, line 41: It's better to propagate errors instead of ignoring

PR #430, opened 2022-10-27
fix: Use proto marshaller instead of amino to derive sign bytes [DEV-1694]

askolesov:


ankurdotb:
Task linked: [DEV-1694 Fix sign bytes derivation](https://app.clickup.com/t/6600954/DEV-1694)

PR #443, opened 2022-11-09
feat: DidDoc versioning [DEV-1892]

askolesov:


ankurdotb:
Task linked: [DEV-1892 Implement DIDDoc versioning on cheqd node](https://app.clickup.com/t/6600954/DEV-1892)

x/did/types/diddoc_metadata.go
```diff
@@ -3,17 +3,16 @@ package types
...
 
-	return Metadata{Created: created, Deactivated: false, VersionId: txHash}
+	return Metadata{Created: created, Updated: created, Deactivated: false, VersionId: version}
 }
 
-func (m *Metadata) Update(ctx sdk.Context) {
+func (m *Metadata) Update(ctx sdk.Context, version string) {
 	m.Updated = ctx.BlockTime().Format(time.RFC3339)
+	m.VersionId = version
 }
```
ankurdotb, line 18: Where's the rest of the metadata relevant such as deactivated, next/previous versions etc? `null` where necessary.

x/did/types/genesis.pb.go
```diff
@@ -83,30 +135,79 @@ func (m *GenesisState) GetFeeParams() *FeeParams {
...
+	0xcc, 0xf4, 0x40, 0xc1, 0xd5, 0x0a, 0xf3, 0xbb, 0xdc, 0x5b, 0x9e, 0x94, 0x1a, 0xa4, 0x55, 0xac,
+	0xf8, 0x2b, 0xf6, 0xe7, 0x9f, 0xa4, 0x5e, 0x7e, 0xef, 0xc6, 0x3f, 0xe1, 0x7c, 0x88, 0x98, 0x16,
+	0x72, 0x2a, 0x27, 0xa6, 0x59, 0x0d, 0x58, 0xa7, 0x1e, 0xef, 0xae, 0x1b, 0x9c, 0x23, 0x5e, 0x5b,
+	0x3a, 0xa9, 0x0d, 0x57, 0xeb, 0x59, 0xf7, 0x6d, 0x2e, 0xd8, 0x6c, 0x2e, 0xd8, 0xc7, 0x5c, 0xb0,
+	0x97, 0x85, 0xa8, 0xcc, 0x16, 0xa2, 0xf2, 0xbe, 0x10, 0x95, 0xbb, 0xc3, 0x91, 0xa2, 0xf1, 0xe3,
+	0x20, 0xcc, 0xf4, 0x24, 0x72, 0x2d, 0xdb, 0x79, 0x94, 0x6b, 0xc0, 0xe8, 0xc9, 0x56, 0x4e, 0xcf,
+	0x05, 0x9a, 0xc1, 0x7f, 0x5b, 0xf9, 0xf1, 0x67, 0x00, 0x00, 0x00, 0xff, 0xff, 0xed, 0xeb, 0xc7,
+	0x91, 0xd1, 0x01, 0x00, 0x00,
+}
+
```
ankurdotb, line 168: What is this and why does it have such a weird name?

ankurdotb (changes requested):
Comments outside of these ones below (since I didn't get to see the protobufs one before merging / looking at everything else in general):

## Resource protobufs

- [ x ] Checksum: Why bytes, not a string? I would prefer just storing the SHA256 checksum rather than in bytes (this is a more commonly understood way of doing it) -> **Bytes formatted on display**
- [ x ] Does it make sense to have a namespace property for resources too? -> **Let resolver handle it**
- [ x ] Unsure if I missed this but I think I did show it, apologies if not: an optional URI property in resources with a repeated map of description and URI. One of these will be filled by default with the current resource URI. -> **Keep it as also known as, let user define a map with description and URI as properties**

## DIDs

### Transactions

- [ ] create-did: Maybe change this to expect a file? Instead of passing verification methods and their private keys one by one. (Also rename to `create-did` instead)
- [ ] Same for update-did

## Tests
- [ ] Looking at the create DID tests, I think it's pretty clear we need perhaps an ADR on update operations and external controllers. The rules implemented for what's allowed vs not for external controllers and updates is up to a DID method to choose, and I'm not a 100% sure some of the current ones are very clear.
- [ ] Probably need more happy/unhappy path scenarios but this be in a different PR
3. `diddoc_service_test`: The services don't have namespace?
4. `diddoc_service_test`: "Base DID is not the same as in ID" <- this doesn't have to be the same as the base DID, no rule says so.
5. Verification method/material tests: No namespace?
6. Verification method tests: "Id does not have expected DID as a base" why does the base need to be the same?
7. Tbh I didn't do all of them since I started seeing a lot of scenarios that were confusing

### Types
1. Do we even need to bother keeping the `types/v1` folder?

## General comments

I skipped a lot of tests near the end since that seems like a beast on its own, and it's probably worth resolving the top-level questions first (since they may be invalid or explainable), and then perhaps revisit the tests. There's also a questions of tests in general vs tests for just versioning.

proto/cheqd/did/v2/tx.proto
```diff
@@ -62,7 +63,8 @@ message MsgUpdateDidDocPayload {
   repeated string key_agreement = 9;
   repeated string also_known_as = 10;
   repeated Service service = 11;
-  string version_id = 12;
+  string previous_version_id = 12;
```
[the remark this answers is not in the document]
ankurdotb: Remove thus

x/did/types/genesis.pb.go
askolesov: This is internals of generated proto. Looks like it keeps bytes of the original `.proto` file. Hard to say why they do it.

proto/cheqd/did/v2/tx.proto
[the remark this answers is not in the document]
askolesov: Removed

askolesov:
I've fixed all notes except tests. Let's review tests in a separate PR.

> Do we even need to bother keeping the types/v1 folder?

That's the whole point of migration. We need to:
* Read old value
* Decode it using old models <- that't where v1 is used
* Transcode to v2 models
* Persist

x/did/types/genesis.pb.go
Eengineer1: It's a way to pass `.proto` field definitions typed to descriptors, as bytes and parse them later, during the decoding process regardless of language.

Eengineer1:
And actually `v1` is an essential part of the migration / upgrade tests PR.

tests/integration/cli/tx.go
```diff
@@ -51,80 +52,90 @@ func Tx(module, tx, from string, txArgs ...string) (sdk.TxResponse, error) {
 	return resp, nil
 }
 
-func CreateDidDoc(payload types.MsgCreateDidDocPayload, signInputs []cli.SignInput, from string) (sdk.TxResponse, error) {
+func CreateDidDoc(tmpDit string, payload types.MsgCreateDidDocPayload, signInputs []cli.SignInput, from string) (sdk.TxResponse, error) {
```
Eengineer1, line 55: @askolesov Simple typo fix.

Eengineer1 (approved):
@askolesov Looks great. Thanks!

Left some concerns regarding the immediate impact on migration / upgrade test suite.

PR #445, opened 2022-11-15
chore: Temporary deactivate buf breaking changes detection

askolesov:
This PR is needed to check if develop branch works aside of breaking change detector


askolesov:
@ankurdotb I'm going to merge this because otherwise all PRs into develop fail because of this check.

PR #449, opened 2022-11-16
fix: Remove cheqd-noded configure command from node [DEV-1902]

benyam7:
I removed the following:
 1.  `configure.go` file  and it's usage inside  `root.go `
 2.  `post_install` method(uses `configure` command) inside `installer.py` and it's usage as well.
 3.  `tests/e2e-complex/interactive-install/add-observer.sh` file
 4. `tests/e2e-complex/upgrade/cosmovisor/add-observer.sh`  file

@askolesov @anikitinDSR  Does this sound okay to you? Or is there anything that I am missing?


ankurdotb:
Task linked: [DEV-1902 Remove cheqd-noded configure command from node & update documentation](https://app.clickup.com/t/6600954/DEV-1902)

installer/installer.py
```diff
@@ -406,43 +406,13 @@ def install(self):
...
-        if self.interviewer.is_setup_needed:
-            self.post_install()
 
         if self.interviewer.init_from_snapshot:
             self.log("Downloading snapshot and extracting archive. This can take a *really* long time...")
             self.download_snapshot()
             self.untar_from_snapshot()
         self.print_success()
 
-    def post_install(self):
```
anikitinDSR, line 418: Let's keep this step, cause such action is still needed. It should be moved to `sed` command calling or something like this. I think, that there is another task for `installer.py`

anikitinDSR (commented):
Please, keep `post_install` method just empty.

installer/installer.py
benyam7: @anikitinDSR Okay,thanks! I kept it empty now.

askolesov: Yes, we can't just get rid of post install here. I would suggest to replace `configure` command with regex based replace

askolesov: @anikitinDSR What do you think about regexps?

askolesov:
Also, I temporarily disabled proto breaking changes check, so we can see if build works or not.

askolesov (commented):
Removal looks good, thanks! The only thing left here is to fix post install script. Take a look at what configure command did and try did the same with regex in python. Please ask me or @anikitinDSR  if you have any questions.

askolesov (approved):

PR #454, opened 2022-11-25
fix: Update  interactive installer for any node changes and test [DEV-1901]

benyam7:


ankurdotb:
Task linked: [DEV-1901 Update  interactive installer for any node changes and test](https://app.clickup.com/t/6600954/DEV-1901)

installer/installer.py
```diff
@@ -428,20 +434,27 @@ def post_install(self):
 
         # Setting up the external_address
         if self.interviewer.external_address:
-            self.exec(f"sudo su -c 'cheqd-noded configure p2p external-address {self.interviewer.external_address}:{self.interviewer.p2p_port}' {DEFAULT_CHEQD_USER}")
-
+            external_address_search_text='external_address = ""'
+            external_address_replace_text='external_address = "{}:{}"'.format(self.interviewer.external_address, self.interviewer.p2p_port)
+            search_and_replace(external_address_search_text, external_address_replace_text, os.path.join(self.cheqd_config_dir, "config.toml"))
```
ankurdotb, line 439: If the user _doesn't_ enter any value here, it should execute line 968-970 and use a live lookup to figure out the IP address of the server and set it. It doesn't seem to do that - or at least, it's not writing it to the config.toml even if it has grabbed that DNS address.

installer/installer.py
```diff
@@ -428,20 +434,30 @@ def post_install(self):
...
+        seeds_search_text = 'seeds = ""'
+        seeds_replace_text= 'seeds = "{}"'.format(seeds)
+        search_and_replace(seeds_search_text, seeds_replace_text, os.path.join(self.cheqd_config_dir, "config.toml"))
+        
         # Setting up the P2P port
-        self.exec(f"sudo su -c 'cheqd-noded configure p2p laddr \"tcp://0.0.0.0:{self.interviewer.p2p_port}\"' {DEFAULT_CHEQD_USER}")
-
+        p2p_laddr_search_text='laddr = "tcp://0.0.0.0:{}"'.format(DEFAULT_P2P_PORT)
+        p2p_laddr_replace_text='laddr = "tcp://0.0.0.0:{}"'.format(self.interviewer.p2p_port)
+        search_and_replace(p2p_laddr_search_text,p2p_laddr_replace_text, os.path.join(self.cheqd_config_dir, "config.toml"))
```
ankurdotb, line 450: Just like `external_address`, shouldn't you check if there's a value entered? Or, if a user enters enter to accept default, is the `p2p_port` value always set to 26656?

installer/installer.py
```diff
@@ -428,20 +434,30 @@ def post_install(self):
...
+        
+        # Setting up the RPC port
+        rpc_laddr_search_text= 'laddr = "tcp://127.0.0.1:{}"'.format(DEFAULT_RPC_PORT)
+        rpc_laddr_replace_text= 'laddr = "tcp://0.0.0.0:{}"'.format(self.interviewer.rpc_port)
+        search_and_replace(rpc_laddr_search_text,rpc_laddr_replace_text, os.path.join(self.cheqd_config_dir, "config.toml"))
+
```
ankurdotb, line 456: By default, the config.toml generated has `tcp://127.0.0.1:26657`. I want this to be replaced in **all** instances by default with `tcp://0.0.0.0:<RPC-PORT>` instead. Then:

1. If the user has entered an RPC port, use that value
2. If the user has NOT entered an RPC port value, use the default value of 26657

ankurdotb (changes requested):
Can you please make these additional changes?

1. Ask user about `log_level` and set default input to `error` (by default, the file has `info`)
2. Ask user about `log_format` and set default input to `json` (by default, file has `plain`)
3. Add a question for `persistent_peers` and add them in (default input blank, file has blank/no value)

Other behaviour to fix:

1. This installer _only_ works with the default version `0.6.9` since it looks assets/files which have a version number without `v` prefix. If I choose one of the non-default options such as `v1.0.0-develop.11`, the installer fails because the asset path will be different. 

```bash
root@installer-test:~# wget -q https://raw.githubusercontent.com/cheqd/cheqd-node/fix-installer/installer/installer.py && sudo python3 installer.py
*********  Latest stable cheqd-noded release version is Name: 0.6.9
*********  List of cheqd-noded releases: 
1) 0.6.9
2) v1.0.0-develop.11
3) v1.0.0-develop.10
4) v1.0.0-develop.9
5) v1.0.0-develop.8
Choose list option number above to select version of cheqd-node to install [default: 1]:
2
Set path for cheqd user's home directory [default: /home/cheqd]:

Do you want to setup a new cheqd-node? (yes/no) [default: yes]:

Select cheqd network to join (testnet/mainnet) [default: mainnet]:
        
*********  INFO: Installing cheqd-node with Cosmovisor allows for automatic unattended upgrades for valid software upgrade proposals.
Install cheqd-noded using Cosmovisor? (yes/no) [default: yes]:

CAUTION: Downloading a snapshot replaces your existing copy of chain data. Usually safe to use this option when doing a fresh installation. Do you want to download a snapshot of the existing chain to speed up node synchronisation? (yes/no) [default: yes]:
no
Provide a moniker for your cheqd-node [default: installer-test]:
ankurs-test
What is the externally-reachable IP address or DNS name for your cheqd-node? [default: Fetch automatically via DNS resolver lookup]: 
142.93.170.2
Specify port for Tendermint RPC [default: 26657]:
26667
Specify port for Tendermint P2P [default: 26656]:
26666
Specify minimum gas price for transactions [default: 25ncheq]:
100ncheq
*********  Downloading cheqd-noded binary...
Reason for failure: No asset found to download for release: v1.0.0-develop.11
Exiting...
Reason for failure: Failed to get cheqd-node binaries from Github
Exiting...
Reason for failure: Unable to install cheqd-node.
Exiting...
```

The reason why this doesn't work anymore is previously, our release versions did NOT have the `v` prefix in the 0.6.x release. We realised that this actually caused issues in being able to call the repo as a golang dependency in other repos, e.g., `did-resolver`. So, starting from v1.x.x, we're going back to adding the `v` prefix. 

The simplest way to deal with this is to check if the asset download URL exists WITHOUT the `v` prefix (since the binary files don't have this), rather than going by the release name. Check the [0.6.9 release](https://github.com/cheqd/cheqd-node/releases/tag/0.6.9) vs [one of the newer v1.0.0-develop releases](https://github.com/cheqd/cheqd-node/releases/tag/v1.0.0-develop.11) to see what I mean in the URL structure for the assets.

installer/installer.py
benyam7: Okay, I see. I tried the command inside VM myself, the problem seems to be not being able to grab it form `stdout` and set it the value to `external_address`. I am looking into it. 👍🏾

installer/installer.py
benyam7: Yep, lots of stuff is making sense now. 👍🏾

installer/installer.py
benyam7: Okay, cool. 👍🏾

installer/installer.py
```diff
@@ -419,29 +427,77 @@ def post_install(self):
...
+        
+        # Setting up persistent peers
+        if self.interviewer.persistent_peers:
+            persistent_peers_search_text='persistent_peers = ""'
+            persistent_peers_replace_text='persistent_peers = "{}"'.format(self.interviewer.persistent_peers)
+            search_and_replace(persistent_peers_search_text,persistent_peers_replace_text, os.path.join(self.cheqd_config_dir, "config.toml"))
+        
+        # Setting up log level
+        if self.interviewer.log_level:
+            log_level_search_text = 'log_level = "info"'
```
ankurdotb, line 484: Just search for `log_level` and replace it, irrespective of current value (this is more robust)

installer/installer.py
```diff
@@ -419,29 +427,77 @@ def post_install(self):
...
+            log_level_replace_text = 'log_level = "{}"'.format(self.interviewer.log_level)
+            search_and_replace(log_level_search_text, log_level_replace_text, os.path.join(self.cheqd_config_dir, "config.toml"))
+        else:
+            log_level_search_text = 'log_level = "info"'
```
ankurdotb, line 488: Same as above

installer/installer.py
```diff
@@ -419,29 +427,77 @@ def post_install(self):
...
+            log_level_replace_text = 'log_level = "{}"'.format(DEFAULT_LOG_LEVEL)
+            search_and_replace(log_level_search_text, log_level_replace_text, os.path.join(self.cheqd_config_dir, "config.toml"))
+
+        # Setting up log format
+        if self.interviewer.log_format:
+            log_format_search_text = 'log_format = "plain"'
```
ankurdotb, line 494: Just search for `log_format` and replace it, regardless of existing value

ankurdotb (changes requested):

ankurdotb (approved):

PR #456, opened 2022-11-30
Add unit tests for UUID

anikitinDSR:


askolesov (commented):
Quick notes:
* Let's add a new method `AddNewResourceVersion` similar to the one in the did module
* Please Add new use case with lover case uuid to make sure it doesn't transform values
* Could you also add test case for version id in did docs?

PR #462, opened 2022-12-02
chore: Bump to Cosmovisor v1.4

benyam7:
This PR bumps Cosmovisor to `v1.3.0` There are not differences between `v1.3.0` and `v1.4.0` except for package name rename. You can see more [here](https://github.com/cosmos/cosmos-sdk/releases?q=cosmovisor&expanded=true). Also, v1.4 doesn't have `tar.gz` file inside the assets.

### DIFF
--- a/installer/installer.py
+++ b/installer/installer.py
@@ -39,7 +39,7 @@ PRINT_PREFIX = "********* "
 DEFAULT_LATEST_COSMOVISOR_VERSION = "v1.3.0"
 COSMOVISOR_BINARY_URL = f"https://github.com/cosmos/cosmos-sdk/releases/download/cosmovisor%2F{DEFAULT_LATEST_COSMOVISOR_VERSION}/cosmovisor-{DEFAULT_LATEST_COSMOVISOR_VERSION}-linux-amd64.tar.gz"
 DEFAULT_USE_COSMOVISOR = "yes"
-DEFAULT_BUMP_COSMOVISOR = "yes"
+DEFAULT_BUMP_COSMOVISOR = "no"
 
 ###############################################################
 ###     				Systemd Config      				###
@@ -416,11 +416,7 @@ class Installer():
 
         if self.interviewer.is_cosmo_needed:
             self.log("Setting up Cosmovisor")
-            self.setup_cosmovisor()
-        
-        if self.interviewer.is_cosmovisor_bump_needed:
-            self.log(f"Bumping Cosmovisor to {DEFAULT_LATEST_COSMOVISOR_VERSION}")
-            self.bump_cosmovisor()
+            self.setup_cosmovisor(self.interviewer.is_cosmovisor_bump_needed)
 
         if not self.interviewer.is_cosmo_needed:
             self.log(f"Moving binary from {self.binary_path} to {DEFAULT_INSTALL_PATH}")
@@ -537,7 +533,7 @@ class Installer():
         except:
             failure_exit(f"Failed to setup {self.cheqd_log_dir} directory")
 
-    def setup_cosmovisor(self):
+    def setup_cosmovisor(self, is_cosmovisor_bump_needed):
         try:
             fname= os.path.basename(COSMOVISOR_BINARY_URL)
             self.exec(f"wget -c {COSMOVISOR_BINARY_URL}")
@@ -548,10 +544,14 @@ class Installer():
             self.remove_safe("CHANGELOG.md")
             self.remove_safe("README.md")
             self.remove_safe("LICENSE")
-            self.mkdir_p(self.cosmovisor_root_dir)
-            self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "genesis"))
-            self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "genesis/bin"))
-            self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "upgrades"))
+            
+            if not is_cosmovisor_bump_needed:
+                self.mkdir_p(self.cosmovisor_root_dir)
+                self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "genesis"))
+                self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "genesis/bin"))
+                self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "upgrades"))
+            else:
+                self.log(f"Bumping Comsovisor binary to {DEFAULT_LATEST_COSMOVISOR_VERSION}")
             if not os.path.exists(os.path.join(DEFAULT_INSTALL_PATH, DEFAULT_COSMOVISOR_BINARY_NAME)):
                 self.log(f"Moving Cosmovisor binary to installation directory")
                 shutil.move("./cosmovisor", DEFAULT_INSTALL_PATH)
@@ -584,52 +584,7 @@ class Installer():
             self.exec(f"chown -R {DEFAULT_CHEQD_USER}:{DEFAULT_CHEQD_USER} {self.cosmovisor_root_dir}")
         except:
             failure_exit(f"Failed to setup Cosmovisor")
-    # This method is exact same with setup_cosmovisor() except it doesn't create existing directories. We could do this inside same method but could lead to coupling.
-    def bump_cosmovisor(self):
-        try:
-            fname= os.path.basename(COSMOVISOR_BINARY_URL)
-            self.exec(f"wget -c {COSMOVISOR_BINARY_URL}")
-            self.exec(f"tar -xzf {fname}")
-            self.remove_safe(fname)
-            
-            # Remove cosmovisor artifacts...
-            self.remove_safe("CHANGELOG.md")
-            self.remove_safe("README.md")
-            self.remove_safe("LICENSE")
-
-            if not os.path.exists(os.path.join(DEFAULT_INSTALL_PATH, DEFAULT_COSMOVISOR_BINARY_NAME)):
-                self.log(f"Moving Cosmovisor binary to installation directory")
-                shutil.move("./cosmovisor", DEFAULT_INSTALL_PATH)
-
-            if not os.path.exists(os.path.join(self.cosmovisor_root_dir, "current")):
-                self.log(f"Creating symlink for current Cosmovisor version")
-                os.symlink(os.path.join(self.cosmovisor_root_dir, "genesis"),
-                        os.path.join(self.cosmovisor_root_dir, "current"))
-
-            self.log(f"Moving binary from {self.binary_path} to {self.cosmovisor_cheqd_bin_path}")
-            self.exec("sudo mv {} {}".format(self.binary_path, self.cosmovisor_cheqd_bin_path))
-            self.exec("sudo chown {} {}".format(f'{DEFAULT_CHEQD_USER}:{DEFAULT_CHEQD_USER}', f'{DEFAULT_INSTALL_PATH}/{DEFAULT_COSMOVISOR_BINARY_NAME}'))
-            self.exec("sudo chmod +x {}".format(f'{DEFAULT_INSTALL_PATH}/{DEFAULT_COSMOVISOR_BINARY_NAME}'))
-
-            if not os.path.exists(os.path.join(DEFAULT_INSTALL_PATH, DEFAULT_BINARY_NAME)):
-                self.log(f"Creating symlink to {self.cosmovisor_cheqd_bin_path}")
-                os.symlink(self.cosmovisor_cheqd_bin_path,
-                        os.path.join(DEFAULT_INSTALL_PATH, DEFAULT_BINARY_NAME))
-
-            if self.interviewer.is_upgrade and \
-                os.path.exists(os.path.join(self.cheqd_data_dir, "upgrade-info.json")):
-
-                self.log(f"Copying upgrade-info.json file to cosmovisor/current/")
-                shutil.copy(os.path.join(self.cheqd_data_dir, "upgrade-info.json"),
-                            os.path.join(self.cosmovisor_root_dir, "current"))
-                self.log(f"Changing owner to {DEFAULT_CHEQD_USER} user")
-                self.exec(f"chown -R {DEFAULT_CHEQD_USER}:{DEFAULT_CHEQD_USER} {self.cosmovisor_root_dir}")
-        
-            self.log(f"Changing directory ownership for Cosmovisor to {DEFAULT_CHEQD_USER} user")
-            self.exec(f"chown -R {DEFAULT_CHEQD_USER}:{DEFAULT_CHEQD_USER} {self.cosmovisor_root_dir}")
-        except:
-            failure_exit(f"Failed to setup Cosmovisor")
-
+    
     def compare_checksum(self, file_path):
         # Set URL for correct checksum file for snapshot
         checksum_url = os.path.join(os.path.dirname(self.interviewer.snapshot_url), "md5sum.txt")
@@ -716,7 +671,7 @@ class Interviewer:
         self._home_dir = home_dir
         self._is_upgrade = False
         self._is_cosmo_needed = True
-        self._is_cosmovisor_bump_needed = True
+        self._is_cosmovisor_bump_needed = False
         self._init_from_snapshot = False
         self._release = None
         self._chain = chain
--- a/installer/installer.py
+++ b/installer/installer.py
@@ -36,8 +36,10 @@ PRINT_PREFIX = "********* "
 ###############################################################
 ###     				Cosmovisor Config      				###
 ###############################################################
-COSMOVISOR_BINARY_URL = "https://github.com/cosmos/cosmos-sdk/releases/download/cosmovisor%2Fv1.1.0/cosmovisor-v1.1.0-linux-amd64.tar.gz"
+DEFAULT_LATEST_COSMOVISOR_VERSION = "v1.3.0"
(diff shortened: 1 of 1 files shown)


ankurdotb:
Task linked: [DEV-1900 Bump to Cosmovisor v1.4](https://app.clickup.com/t/6600954/DEV-1900)

installer/installer.py
```diff
@@ -36,8 +36,10 @@
 ###############################################################
 ###     				Cosmovisor Config      				###
 ###############################################################
-COSMOVISOR_BINARY_URL = "https://github.com/cosmos/cosmos-sdk/releases/download/cosmovisor%2Fv1.1.0/cosmovisor-v1.1.0-linux-amd64.tar.gz"
+DEFAULT_LATEST_COSMOVISOR_VERSION = "v1.3.0"
+COSMOVISOR_BINARY_URL = f"https://github.com/cosmos/cosmos-sdk/releases/download/cosmovisor%2F{DEFAULT_LATEST_COSMOVISOR_VERSION}/cosmovisor-{DEFAULT_LATEST_COSMOVISOR_VERSION}-linux-amd64.tar.gz"
```
ankurdotb, line 40: Just realised that this (and our previous version) would _always_ download the linux_amd64 version of Cosmovisor. (`x86_64` is what we call our `amd64` build.) At a later step in the process, we determine which platform architecture the installer is running and accordingly download the appropriate binary.

We should do something similar here. If the user is on a Linux amd64/x86_64 systems, we should download the `linux_amd64` binary. Otherwise, when the user is running Linux ARM64, we should download the appropriate binary for Cosmvisor.

Note that Cosmovisor also generates assets for MacOS/Darwin, Windows etc - we don't need to account for these, since cheqd-node is not compiled for these platforms.

installer/installer.py
```diff
@@ -1114,7 +1138,11 @@ def install_steps():
 
     # Steps to execute if upgrading existing node
     def upgrade_steps():
-        interviewer.ask_for_cosmovisor()
+        # if cosmovisor is not installed 
+        if not os.path.exists(os.path.join(DEFAULT_INSTALL_PATH, DEFAULT_COSMOVISOR_BINARY_NAME)):
+            interviewer.ask_for_cosmovisor()
+        else:
+            interviewer.ask_for_cosmovisor_bump()
```
ankurdotb, line 1145: If there's a Cosmovisor binary installed, this will _always_ ask to bump - even if the user already has Cosmovisor 1.3 installed. (Remember that someone might run the upgrade more than once.)

We don't need to do this. You can actually check the version of Cosmovisor running using the output of `cosmovisor version`. **Only** ask about bumping Cosmvisor in case the version is lower than the default Cosmovisor version set at the top. Otherwise, no need to ask this question.

Same logic applies when installing "from scratch". Even if there's a node already installed, remember the installer gives the option to install from scratch. So we should test whether in the install from scratch path, the new version of Cosmovisor needs to be installed or not.

installer/installer.py
```diff
@@ -36,8 +36,10 @@
 ###############################################################
 ###     				Cosmovisor Config      				###
 ###############################################################
-COSMOVISOR_BINARY_URL = "https://github.com/cosmos/cosmos-sdk/releases/download/cosmovisor%2Fv1.1.0/cosmovisor-v1.1.0-linux-amd64.tar.gz"
+DEFAULT_LATEST_COSMOVISOR_VERSION = "v1.3.0"
+COSMOVISOR_BINARY_URL = f"https://github.com/cosmos/cosmos-sdk/releases/download/cosmovisor%2F{DEFAULT_LATEST_COSMOVISOR_VERSION}/cosmovisor-{DEFAULT_LATEST_COSMOVISOR_VERSION}-linux-amd64.tar.gz"
 DEFAULT_USE_COSMOVISOR = "yes"
+DEFAULT_BUMP_COSMOVISOR = "no"
```
ankurdotb, line 42: Default should be yes

installer/installer.py
```diff
@@ -665,6 +671,7 @@ def __init__(self,
         self._home_dir = home_dir
         self._is_upgrade = False
         self._is_cosmo_needed = True
+        self._is_cosmovisor_bump_needed = False
```
ankurdotb, line 674: Default should be yes

installer/installer.py
```diff
@@ -542,10 +544,14 @@ def setup_cosmovisor(self):
             self.remove_safe("CHANGELOG.md")
             self.remove_safe("README.md")
             self.remove_safe("LICENSE")
-            self.mkdir_p(self.cosmovisor_root_dir)
-            self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "genesis"))
-            self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "genesis/bin"))
-            self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "upgrades"))
+            
+            if not is_cosmovisor_bump_needed:
```
ankurdotb, line 548: Remember that `setup_cosmovisor` is **also** used for new/setup from scratch. Don't assume that this logic needs to execute only if a bump is needed.

I'd keep the logic of `setup_cosmovisor` exactly as it is now since it covers the install from scratch, or in upgrade scenario where the user chooses yes to install from scratch use case. Split out the function for whether a Cosmovisor bump is needed to its own function, that way you don't overload this function.

installer/installer.py
```diff
@@ -542,10 +544,14 @@ def setup_cosmovisor(self):
...
+                self.mkdir_p(self.cosmovisor_root_dir)
+                self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "genesis"))
+                self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "genesis/bin"))
+                self.mkdir_p(os.path.join(self.cosmovisor_root_dir, "upgrades"))
+            else:
+                self.log(f"Bumping Comsovisor binary to {DEFAULT_LATEST_COSMOVISOR_VERSION}")
```
ankurdotb, line 554: As far as I understand, this will only log the message and do absolutely nothing else (which is incorrect)

installer/installer.py
```diff
@@ -542,10 +544,14 @@ def setup_cosmovisor(self):
...
             if not os.path.exists(os.path.join(DEFAULT_INSTALL_PATH, DEFAULT_COSMOVISOR_BINARY_NAME)):
```
ankurdotb, line 555: If Cosmovisor is already installed at v1.1.0 and the user has asked to bump it, this logic will never be executed (if I'm right)

ankurdotb (changes requested):
See ClickUp for comments

installer/installer.py
```diff
@@ -1114,7 +1211,9 @@ def install_steps():
 
     # Steps to execute if upgrading existing node
     def upgrade_steps():
-        interviewer.ask_for_cosmovisor()
+        # if cosmovisor is not installed 
+        if os.path.exists(os.path.join(DEFAULT_INSTALL_PATH, DEFAULT_COSMOVISOR_BINARY_NAME)):
```
ankurdotb, line 1215: Move logic to its own function, don't introduce this into the main/Interviewer

installer/installer.py
```diff
@@ -994,6 +1068,29 @@ def ask_for_cosmovisor(self):
             self.is_cosmo_needed = False
         else:
             failure_exit(f"Invalid input provided during installation.")
+    
+    def ask_for_cosmovisor_bump(self):
+        self.exec("echo 'export DAEMON_NAME=cheqd-noded' >> ~/.bashrc")
+        self.exec(f"echo 'export DAEMON_HOME={self.home_dir}/.cheqdnode' >> ~/.bashrc")
+        self.exec(f"echo 'export DAEMON_DATA_BACKUP_DIR={self.home_dir}/.cheqdnode' >> ~/.bashrc")
```
ankurdotb, line 1075: 1. The installer might be executing as ANY user, and therefore you need to ensure it's put in THAT user's environment variables, but also in the cheqd user's environment variables.
2. Use Python-native libraries to write to the file. You don't actually know whether these files are writeable or not. https://stackoverflow.com/questions/17657686/is-it-possible-to-set-an-environment-variable-from-python-permanently

installer/installer.py
```diff
@@ -994,6 +1068,29 @@ def ask_for_cosmovisor(self):
...
+        std_out = """{}""".format(self.exec("cosmovisor version 2> /dev/null || echo false").stdout)     
+        arr = std_out.split()
+        current_version = None
```
ankurdotb, line 1078: 1. You can send the output to a file
2. You can read the file, and just read the first line
3. Use regex pattern matching to extract the last word https://www.tutorialspoint.com/How-to-match-at-the-end-of-string-in-python-using-Regular-Expression
4. If present, ditch the v tag (regex already does this) `(\d+)(\.\d+)?(\.\d+)?$`

installer/installer.py
```diff
@@ -994,6 +1068,29 @@ def ask_for_cosmovisor(self):
...
+        for i in range (0, len(arr)):
+            if arr[i].lower() == 'version:':
+                # set current version as well as replace if v prefix there
+                current_version = arr[i+1].split("\\")[0].replace('v','')
+                break
+        # if current version is lower than latest ask for upgrade
```
ankurdotb, line 1084: 1. Have an is_cosmovisor exists, where you check is it installed, where is it installed, and what version it is
2. If it's installed and older version, ask for bump

installer/installer.py
```diff
@@ -96,6 +103,10 @@ def search_and_replace(search_text, replace_text, file_path):
             
     file.close()
 
+def version_tuple(v):
+    return tuple(map(int, (v.split("."))))
```
ankurdotb, line 107: Just strip `v` if present and compare in `is_cosmovisor_bump_needed' using string match

installer/installer.py
```diff
@@ -412,9 +423,13 @@ def install(self):
         self.prepare_directory_tree()
         self.setup_system_configs()
 
-        if self.interviewer.is_cosmo_needed:
+        if self.interviewer.is_cosmo_needed and self.interviewer.is_cosmovisor_bump_needed:
```
ankurdotb, line 426: This condition might not be needed since the interviewer handles whether cosmo needed/bump needed is required.

installer/installer.py
ankurdotb: This should be done in bump_cosmovisor, NOT in ask_for_cosmovisor

benyam7: @ankurdotb Regarding checking file permission I tried this function, but it always returns false for both these 2 file paths I pass.
```is_file_writable("/home/cheqd/.bashrc")```
```is_file_writable(os.path.expanduser("~/.bashrc"))```
```python3
def is_file_writable(self, path_to_file):
        if os.path.exists(path_to_file) and os.path.isfile(path_to_file): 
            return os.access(path_to_file, os.W_OK)
        return False
```
And since the `glibc` issue is not related to that, I didn't want to so much time on it.

installer/installer.py
```diff
@@ -576,9 +671,51 @@ def setup_cosmovisor(self):
...
+    def set_env_vars(self, env_var_name, env_var_value):
+        if not self.check_if_env_var_already_set(env_var_name) and not self.is_default_shell_fish():
+            self.write_to_bashrc(env_var_name, env_var_value) # write to current user
+            self.write_to_bashrc(env_var_name, env_var_value, True) # write to cheqd user .bashrc
+        elif not self.check_if_env_var_already_set(env_var_name) and self.is_default_shell_fish():
+            self.exec(f'set -Uxg {env_var_name} {env_var_value}')
+        else:
+            self.log(f"ENV var {env_var_name} already set")
+
+    def write_to_bashrc(self, env_var_name, env_var_value, is_user_cheqd=False):
```
ankurdotb, line 687: `is_user_cheqd` will always be False, but I assume this is work-in-progress. Ideally, we want to set it for the `cheqd` user AND whichever user is executing the script. This will ensure a start from _either_ user refers to the same config (for example, `cheqd-noded start` right now results in very different things if done as root vs done as cheqd).

installer/installer.py
```diff
@@ -559,12 +601,65 @@ def setup_cosmovisor(self):
...
+            self.set_env_vars("DAEMON_HOME",f"{self.interviewer.home_dir}/.cheqdnode")
+            self.set_env_vars("DAEMON_DATA_BACKUP_DIR",f"{self.interviewer.home_dir}/.cheqdnode")
+        except:
+            failure_exit(f"Failed to setup Cosmovisor")
+    
+    def bump_cosmovisor(self):
+        try:
+            self.stop_cosmovisor_systemd(DEFAULT_COSMOVISOR_SERVICE_NAME)
+            fname= os.path.basename(self.cosmovisor_download_url)
+            self.exec(f"wget -c {self.cosmovisor_download_url}")
```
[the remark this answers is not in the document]
benyam7: Resolved. We do the same thing inside `get_binary()` but with a different flags. Also, now `get_binary()` untars different versions using different way so  I kept it as it is.

installer/installer.py
```diff
@@ -32,15 +32,22 @@
 DEFAULT_CHEQD_USER = "cheqd"
 DEFAULT_BINARY_NAME = "cheqd-noded"
 DEFAULT_COSMOVISOR_BINARY_NAME = "cosmovisor"
-DEFAULT_CHAINS = ['testnet', 'mainnet']
-DEFAULT_CHAIN = "mainnet"
+DEFAULT_CHAINS = ['cheqd-testnet-4', 'cheqd-mainnet-1']
```
ankurdotb, line 35: Let's discuss whether we ask them for just mainnet/testnet and set chain ID based on that.

installer/installer.py
```diff
@@ -32,15 +32,22 @@
...
-COSMOVISOR_BINARY_URL = "https://github.com/cosmos/cosmos-sdk/releases/download/cosmovisor%2Fv1.1.0/cosmovisor-v1.1.0-linux-amd64.tar.gz"
+DEFAULT_LATEST_COSMOVISOR_VERSION = "v1.2.0"
+COSMOVISOR_BINARY_URL = "https://github.com/cosmos/cosmos-sdk/releases/download/cosmovisor%2F{}/cosmovisor-{}-linux-{}.tar.gz"
 DEFAULT_USE_COSMOVISOR = "yes"
+DEFAULT_BUMP_COSMOVISOR = "yes"
+DEFAULT_DAEMON_ALLOW_DOWNLOAD_BINARIES = "true"
+DEFAULT_DAEMON_RESTART_AFTER_UPGRADE = "true"
+DEFAULT_DAEMON_POLL_INTERVAL = "300s"
+DEFAULT_UNSAFE_SKIP_BACKUP = "true"
+DEFAULT_CONTINUE_DIRTY_COSMOVISOR_STATE = False
```
ankurdotb, line 50: This logic branch is not needed

installer/installer.py
```diff
@@ -267,7 +292,10 @@ def get_binary(self):
         try:
             self.exec(f"wget -c {binary_url}")
             if fname.find(".tar.gz") != -1:
-                self.exec(f"tar -xzf {fname} -C .")
+                if self.version.replace('v', '') == '0.6.9':
```
ankurdotb, line 295: Special case for v0.6.9 is probably not needed any more because latest in that release family is v0.6.10 and URL structure is same as all other version

ankurdotb (changes requested):
1. In workflow, might need to use ENV in workflow to override the file path for systemctl files to use the copy of service files in same branch
2. Add a test for node-standalone install (currently not tested in pipeline).

installer/installer.py
```diff
@@ -32,15 +32,22 @@
...
-COSMOVISOR_BINARY_URL = "https://github.com/cosmos/cosmos-sdk/releases/download/cosmovisor%2Fv1.1.0/cosmovisor-v1.1.0-linux-amd64.tar.gz"
+DEFAULT_LATEST_COSMOVISOR_VERSION = "v1.2.0"
+COSMOVISOR_BINARY_URL = "https://github.com/cosmos/cosmos-sdk/releases/download/cosmovisor%2F{}/cosmovisor-{}-linux-{}.tar.gz"
 DEFAULT_USE_COSMOVISOR = "yes"
+DEFAULT_BUMP_COSMOVISOR = "yes"
+DEFAULT_DAEMON_ALLOW_DOWNLOAD_BINARIES = "true"
+DEFAULT_DAEMON_RESTART_AFTER_UPGRADE = "true"
+DEFAULT_DAEMON_POLL_INTERVAL = "300s"
+DEFAULT_UNSAFE_SKIP_BACKUP = True
+DEFAULT_CONTINUE_DIRTY_COSMOVISOR_STATE = False
```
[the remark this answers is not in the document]
abdulla-ashurov: Outdated. In installer.py I have already changed `DEFAULT_UNSAFE_SKIP_BACKUP = True` (boolean) to `DEFAULT_UNSAFE_SKIP_BACKUP = "true"` (string)

installer/installer.py
abdulla-ashurov: Done! ✅

installer/installer.py
abdulla-ashurov: Done! ✅

installer/installer.py
abdulla-ashurov: Done! ✅

abdulla-ashurov:
> 1. In workflow, might need to use ENV in workflow to override the file path for systemctl files to use the copy of service files in same branch
> 2. Add a test for node-standalone install (currently not tested in pipeline).

@ankurdotb
DONE! ✅

installer/installer.py
```diff
@@ -4,89 +4,128 @@
 ###############################################################
 ###     		    Python package imports      			###
 ###############################################################
-import datetime
-import os
-import subprocess
-import sys
-import urllib.error
 from pathlib import Path
```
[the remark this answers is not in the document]
anikitinDSR: Yes, we are using it for touching the file.

ankurdotb (approved):

PR #466, opened 2022-12-05
feat(swagger): Autogenerate swagger for cheqd-node protobufs [DEV-1661]

benyam7:


ankurdotb:
Task linked: [DEV-1661 Autogenerate swagger for cheqd-node protobufs](https://app.clickup.com/t/6600954/DEV-1661)

make/proto.mk
```diff
@@ -27,41 +27,41 @@ DOCKER_BUF := docker run -v $(shell pwd):/workspace --workdir /workspace bufbuil
...
 
-# proto-lint:
-# 	@$(DOCKER_BUF) lint --error-format=json
+proto-lint:
+	@$(DOCKER_BUF) lint --error-format=json
 
-# proto-check-breaking:
-# 	@$(DOCKER_BUF) breaking --against https://github.com/cheqd/cheqd-node.git#branch=main
+proto-check-breaking:
+	@$(DOCKER_BUF) breaking --against https://github.com/cheqd/cheqd-node.git#branch=main
```
ankurdotb, line 34: Not needed, this is already in the workflow file

make/proto.mk
```diff
@@ -27,41 +27,41 @@ DOCKER_BUF := docker run -v $(shell pwd):/workspace --workdir /workspace bufbuil
...
 
-# GOGO_PROTO_URL           = https://raw.githubusercontent.com/regen-network/protobuf/cosmos
-# GOOGLE_PROTO_URL         = https://raw.githubusercontent.com/googleapis/googleapis/master
-# COSMOS_PROTO_URL         = https://raw.githubusercontent.com/cosmos/cosmos-sdk/v0.45.4/proto/cosmos
-# COSMOS_ORM_PROTO_URL     = https://raw.githubusercontent.com/cosmos/cosmos-sdk/orm/v1.0.0-alpha.10/proto/cosmos
+GOGO_PROTO_URL           = https://raw.githubusercontent.com/regen-network/protobuf/cosmos
+GOOGLE_PROTO_URL         = https://raw.githubusercontent.com/googleapis/googleapis/master
+COSMOS_PROTO_URL         = https://raw.githubusercontent.com/cosmos/cosmos-sdk/v0.45.4/proto/cosmos
```
ankurdotb, line 38: If required (and I don't think it is), make this match Cosmos SDK version in go.mod

proto/buf.gen.swagger.yaml
```diff
@@ -4,5 +4,5 @@ version: v1
 
 plugins:
   - name: swagger
-    out: ../tmp-swagger-gen
+    out: ../app/client/docs
```
ankurdotb, line 7: This path is valid on regen but not on our network. Maybe publish this on a top level folder called `api`

scripts/protoc-swagger-gen.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# find all proto directories
+proto_dirs=$(find ./cheqd -path -prune -o -name '*.proto' -print0 | xargs -0 -n1 dirname | sort | uniq)
+
+# loop through proto directories
+for dir in $proto_dirs; do
+  # generate swagger files (filter query files)
+  query_file=$(find "${dir}" -maxdepth 1 \( -name 'query.proto' \))
+  if [[ -n "$query_file" ]]; then
+    buf generate --template buf.gen.swagger.yaml "$query_file"
```
ankurdotb, line 16: THIS is the main thing - it's the bit that looks at our own proto directories and generates does buf generate. Check if there's a buf generate plugin for creating Swager files, which might be the easiest way to do it.

make/proto.mk
```diff
@@ -27,41 +27,41 @@ DOCKER_BUF := docker run -v $(shell pwd):/workspace --workdir /workspace bufbuil
...
 
-# 	@mkdir -p $(COSMOS_PROTO_TYPES)/base/query/v1beta1/
-# 	@curl -sSL $(COSMOS_PROTO_URL)/base/query/v1beta1/pagination.proto > $(COSMOS_PROTO_TYPES)/base/query/v1beta1/pagination.proto
+	@mkdir -p $(COSMOS_PROTO_TYPES)/base/query/v1beta1/
+	@curl -sSL $(COSMOS_PROTO_URL)/base/query/v1beta1/pagination.proto > $(COSMOS_PROTO_TYPES)/base/query/v1beta1/pagination.proto
 
-# 	@mkdir -p $(COSMOS_PROTO_TYPES)/orm/v1alpha1/
-# 	@curl -sSL $(COSMOS_ORM_PROTO_URL)/orm/v1alpha1/orm.proto > $(COSMOS_PROTO_TYPES)/orm/v1alpha1/orm.proto
+	@mkdir -p $(COSMOS_PROTO_TYPES)/orm/v1alpha1/
+	@curl -sSL $(COSMOS_ORM_PROTO_URL)/orm/v1alpha1/orm.proto > $(COSMOS_PROTO_TYPES)/orm/v1alpha1/orm.proto
```
ankurdotb, line 60: Correct me if I'm wrong but this will generate for base Cosmos types and we don't care about that.

scripts/generate_swagger_docs.sh
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env bash
```
ankurdotb, line 1: Unsure if you need this file

ankurdotb (changes requested):
None of the Swagger generation has actually been run. We only care about generating API docs specifically for the query/transactions we have in the `proto` folder. 

I suggest you look at the following two different approaches:

1. `did-resolver` uses comments/annotation in source with a Golang package called `swag` to generate documents. This could work, but requires editing source code.
2. Alternatively, I think we already have grpc-web support in this natively as well as the Protobufs. Try just invoking just the shell script with buf generate and that itself might give valid output. (While it's not used for Swagger doc generation, a similar technique using buf generate with a remote plugin is used in the `ts-proto` repo.)

make/proto.mk
benyam7: Yep, not required. I was able to generate the docs with these.

make/proto.mk
benyam7: yep, no need for that as well.

benyam7:
@ankurdotb we can gitignore generated swagger files. But, I have seen  Cosmos-sdk keeps them but Regen-ledger ignore them. Open for discussion.

benyam7:
@ankurdotb Also, the lint is failing because it is assuming `config.json` as `swagger.json` aka the one that contains everything for swagger doc generation. But, `config.json` is just a tool used to combine the two separate swagger generated modules aka v2/did and v2/resources by `swagger-combine`. So, ignoring the `config.json` file is the way to go.

proto/cheqd/did/v2/tx.proto
```diff
@@ -3,78 +3,115 @@ syntax = "proto3";
...
-  repeated string key_agreement = 9;
-  repeated string also_known_as = 10;
-  repeated Service service = 11;
-  string version_id = 12;
+  // Contents of the DID Document to be created
+  DidDoc did_doc = 1 [
+    (gogoproto.nullable) = false,
+    (gogoproto.embed) = true,
+    (gogoproto.jsontag) = "didDocument"
+  ];
```
Eengineer1, line 68: I don't think it makes sense to add extra nested layer on top.
It'll affect down all initialisation and require an extra step on building the message (currently it's 2 steps max, will be 3).

This is a breaking change on message server as well.

proto/cheqd/did/v2/tx.proto
```diff
@@ -3,78 +3,115 @@ syntax = "proto3";
...
-  repeated string capability_delegation = 8;
-  repeated string key_agreement = 9;
-  repeated string also_known_as = 10;
-  repeated Service service = 11;
+  // Contents of the DID Document to be updated
+  DidDoc did_doc = 1 [
+    (gogoproto.nullable) = false,
+    (gogoproto.embed) = true,
+    (gogoproto.jsontag) = "didDocument"
+  ];
```
Eengineer1, line 89: Same logic applies to this as well.

Eengineer1 (approved):
Looks good.

Non-breaking todo: 
- Fix linting issue with `api/docs/config.yaml` associated with `swagger-combine`.

anikitinDSR (approved):

PR #480, opened 2022-12-14
test: Add the ability to read env variables

anikitinDSR:
Environment variables:
- UPGRADE_NAME - uses to change the upgrade name
- PROPOSAL_ID      - proposal id
- CHEQD_NODE_HOME - allows to change the directory with keys. By default it compiles the path from "docker/localnet/network-config" + operator name. Like "docker/localnet/network-config/validator-0". In case of copying keys to "/home/cheqd/.cheqdnode" this value can be set as CHEQD_NODE_HOME.
- DEBUG - in case of setting "DEBUG=true" will print the whole string of command, like "cheqd-noded tx gov vote 2 yes --from operator-3 --gas auto --gas-adjustment 1.8 --gas-prices 25ncheq --keyring-backend test --chain-id cheqd -y --home ../../../docker/localnet/network-config/validator-3". Works for docker and local executions.

### DIFF
--- a/tests/upgrade/integration/cli/config.go
+++ b/tests/upgrade/integration/cli/config.go
@@ -48,4 +48,6 @@ var (
 		"--chain-id", integrationnetwork.CHAIN_ID,
 		"--output", OUTPUT_FORMAT,
 	}
+	// TODO: UGLY THING, needs to be refactored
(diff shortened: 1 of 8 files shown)


Tweeddalex:
@anikitinDSR what is the status of this PR, is it now deprecated?

PR #498, opened 2023-01-04
ics23 v0.9.0 and ibc v6.1.0

faddat:
This PR allows cheqd to use ics23 v0.9.0 and ibc-go v6.1.0.

It removes the dragonberry replace, which is no longer needed in sdk v0.46.7+


ankurdotb:
@faddat We've fixed this under #500 and merged it to our develop branch as a release candidate :) We've got some minor fixes to run through next week but thanks for backing up the move to IBC v6

PR #506, opened 2023-01-10
fix(ibc): Fixed router seal & legacy bash tests latest working state [DEV-2098]

Eengineer1:
- Fixes IBC router seal.
- Brings legacy IBC tests to working state.
- Updates legacy IBC tests to latest relevant versions.


ankurdotb (commented):
Fixed and moved files to a different structure. Also, dropped the need to re-build images since this is extremely redundant for our use case, as far as I can see.

ankurdotb:
Task linked: [DEV-2098 Execute manual IBC test](https://app.clickup.com/t/6600954/DEV-2098)

Eengineer1:
> Fixed and moved files to a different structure. Also, dropped the need to re-build images since this is extremely redundant for our use case, as far as I can see.

Only use case I see for keeping image builds is bootstrapping it locally out of the box, without a docker login to registry.

PR #511, opened 2023-01-16
fix: Fix for converting old time to timestamp

anikitinDSR:


Eengineer1 (approved):
Fixed cases look good.
Approved.

PR #524, opened 2023-01-24
build(tendermint): Switch to Informal Systems Tendermint fork for Cosmos SDK v0.46.8

ankurdotb:


Eengineer1 (dismissed):
Dependencies look good.
Approved.

PR #553, opened 2023-02-27
build: Add publish API sub-package for cheqd-node [DEV-2232]

abdulla-ashurov:
- added publish API sub-package for cheqd-node


ankurdotb:
Task linked: [DEV-2232 Change workflows in cheqd node and add publish API sub-package for cheqd-node](https://app.clickup.com/t/6600954/DEV-2232)

.github/workflows/dispatch.yml
```diff
@@ -30,3 +30,10 @@ jobs:
     if: ${{ github.ref_protected == true }}
     uses: ./.github/workflows/release.yml
     secrets: inherit
+  
+  call-release-api:
+    name: "Release API"
+    needs: call-release
+    uses: ./.github/workflows/release-api.yml
```
anikitinDSR, line 37: This one is not compatible with `on.push` as you are using below in `release-api.yml`. Please, pick only one of calling method. I assume, it should be independent from releasing process and reflects only changes in `api/` directory.

anikitinDSR (commented):
Maybe instead of detecting all the changes in `api/` directory it worse to detect only really significant changes in `go` files?

PR #562, opened 2023-03-07
feat: Fix multibase && fix base58 encoding in debug cmd

DaevMithran:
- Fixed multibase encoding by adding prefix
  -  Should I use the helper function `GenerateEd25519VerificationKey2020VerificationMaterial` instead of the repeated code here?
- Added base58 encoding

- Should we move the `pubkey-base64-to-jwk` command into encoding as well?, currently it's a subcommand of ed25519


anikitinDSR:
> 

You could use `GenerateEd25519VerificationKey2020VerificationMaterial` in case of representing 32-symbol base58 string to multibase. If the input from base64 is base58 string you can maybe encode bytes after base64 decoding into string and use such function. What is suppose to be inside base64?

Eengineer1 (approved):
Looks good.

Other than that, enhanced descriptions and namings.
Thanks!

PR #639, opened 2023-08-30
fix: makefile LDFLAGS

whalelephant:
This allows to build static binary without deps - to be used in Heighliner.

[agouin](https://github.com/agouin) identified `LDFLAGS` was missing in  https://github.com/strangelove-ventures/heighliner/issues/166


Eengineer1 (approved):
Looks good + straightforward.

PR #685, opened 2024-01-25
fix: Add msg signature normalisation for deactivate did doc msg

lampkin-diet:


Eengineer1 (approved):
Looks good, non-breaking as well.

GoReleaser is failing since it's pulling a direct dependency requiring Go v1.20 at least.

That will be addressed in the scope of the upcoming release, firstly bumping to Go v1.21 as the latest stable at the time of writing.

PR #690, opened 2024-02-07
build(deps): bump cosmos-sdk to v047

atheeshp:
Closes: #686

PR #694, opened 2024-02-20
fix: failing tests

atheeshp:


tests/upgrade/integration/v2/generated/post/query - module-version-map/v2.json
```diff
@@ -23,11 +23,11 @@
             },
```
Eengineer1, line 23: I've highlighted the relevant places whence the `consensus` module is missing within #690.

Let's add it here as well, as it wasn't registered at all, with the test giving false positives.

Eengineer1: Its version should be `1`.

PR #708, opened 2024-03-07
fix: state sync fixes using upgrade handler

deepan95dev:
This fixes the state sync issue while upgrading sdk version. with below error

`snapshot restoration failed: app version mismatch. Expected: 0, got: 2`

> Solution: In upgrade handler, consensus param App version is set to 2 which is required by cometbft ABCI App version check [here](https://github.com/cometbft/cometbft/blob/cce2e5da5b7f54aed1fe6dd36c14404ad941d714/statesync/syncer.go#L493)


Eengineer1 (approved):
Looks straightforward.

Approved.

Nice work!

Issue #774, opened 2024-08-23
[feature]: integrate x/feemarket module in app


akhilkumarpilli:
### What is the feature you'd like to see?

Integrate x/feegrant module in app from [cheqd/feemarket](https://github.com/cheqd/feemarket/tree/sdk-47) which is forked from `skip-mev`

### Are there any examples of where similar features are available?

 [cheqd/feemarket](https://github.com/cheqd/feemarket/tree/sdk-47) 

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #778, opened 2024-08-23
feat!: integrate feemarket

vishal-kanna:
closes: #774


Eengineer1:
@vishal-kanna Same logic as #779  applies here.

PR #783, opened 2024-09-10
feat: Added MsgBurn per signer

vishal-kanna:


x/did/client/cli/burn_cli.go
```diff
@@ -0,0 +1,44 @@
...
+
+			txf, err := tx.NewFactoryCLI(clientCtx, cmd.Flags())
+			if err != nil {
+				return err
+			}
+
+			amount, err := sdk.ParseCoinNormalized(args[0])
+			if err != nil {
+				return err
+			}
```
Eengineer1, line 31: We should parse the coin list holistically, rather than collecting the first element of the array + validating.

The intended behaviour here is to burn any supported denomination present within the list, so technically it can be more than one coin list elements included.

x/did/keeper/msg_server.go
```diff
@@ -149,3 +153,45 @@ func VerifyAllSignersHaveAtLeastOneValidSignature(k *Keeper, ctx *sdk.Context, i
...
+	}
+
+	bondDenom := k.stakingKeeper.BondDenom(sdkCtx)
+	denoms := msg.Amount.Denoms()
+	if len(denoms) != 0 {
+		err := ValidateDenom(denoms[0], bondDenom)
+		if err != nil {
+			return nil, err
+		}
+	}
```
Eengineer1, line 172: Similarly, there might exist more than one records, so all _must_ be validated accordingly rather than just the first list element.

x/did/types/tx_msg_burn.go
```diff
@@ -0,0 +1,49 @@
...
+	}
+}
+
+// NewMsgBurn creates a message to burn tokens
+func NewMsgBurnFrom(amount sdk.Coins, fromAddress string) *MsgBurn {
+	return &MsgBurn{
+		Amount:      amount,
+		FromAddress: fromAddress,
+	}
+}
```
Eengineer1, line 25: This is unnecessary, as only the signer can burn from their respective account.

Eengineer1 (changes requested):
The core work looks good.

Please, address + resolve the relevant validation logic remarks, before that's ready to merge.

Eengineer1 (approved):
Looks well polished.

Approved.

PR #788, opened 2024-09-26
bump ubuntu to 22.04

atheeshp:


Eengineer1:
@filipdjokic Could you please address similarly with the v2.x release line?

The issues are identical.

filipdjokic:
@Eengineer1 Opened a PR to apply changes on 2.x release.

Also, I would rather set the version to `ubuntu-24.04` instead of `ubuntu-latest`, since it's not yet marked as the latest - see this [PR comment](https://github.com/actions/runner-images/issues/10636#issuecomment-2401625371).

If you check the workflows execution logs after the change, you'll see that it ran on Ubuntu 22.04 LTS.

Eengineer1:
> @Eengineer1 Opened a PR to apply changes on 2.x release.
> 
> Also, I would rather set the version to `ubuntu-24.04` instead of `ubuntu-latest`, since it's not yet marked as the latest - see this [PR comment](https://github.com/actions/runner-images/issues/10636#issuecomment-2401625371).
> 
> If you check the workflows execution logs after the change, you'll see that it ran on Ubuntu 22.04 LTS.

That's correctly spotted, but `ubuntu-24.04` is against latest Super Linter's valid GH actions values, hence the explicit switch to `ubuntu-latest`.

image

ankurdotb:
@filipdjokic Will one your your PRs supersede this? Will you cherry-pick from `release/2.x` to here?

PR #805, opened 2024-10-16
fix: Added explicit did module account permission setter within upgrade handler

vishal-kanna:


Eengineer1 (approved):
The explicit workaround for setting module account permissions looks good.

Approved.

PR #839, opened 2025-01-14
docs: Add v3.1.3 binaries.json file [DEV-4669]

filipdjokic:


ankurdotb:
Task linked: [DEV-4669 Add upgrades.json file](https://app.clickup.com/t/6600954/DEV-4669)

networks/testnet/upgrades/upgrade-v3.1.3.json
ankurdotb, line 1: Rename to upgrade-v3.1.json in case we need to update. Upgrade handler only uses the major and minor version, not the patch version

ankurdotb (changes requested):

ankurdotb (approved):

PR #886, opened 2025-05-19
build: Protobuf lockfile updates + run Make commands [DEV-5011]

ankurdotb:
Our goreleaser file lists the following commands to execute hooks:

    - make clean
    - make tidy
    - make proto-gen
    - make swagger

In practice, when this runs on CI, this should not result in a change because any changes have already been executed locally and checked in, similar to checking in lockfiles.

Of particular interest are `make tidy` (which can change the go.mod, go.sum files) and `make proto-gen`. The latter will only create a difference _if_ there's been a change in the `buf.lock` file to update to a newer Protobuf definition.

In our case, the `buf.lock` file was updated during the last major release, whereas the upstream Buf repository has changes since - especially with differences in `buf.build/cosmos/cosmos-sdk` as it's gone from Cosmos SDK v0.47 to Cosmos SDK v0.50.

Sequence of commands executed to get to this stage in the PR:

```
cd proto
buf mod update
cd ..
make clean
make tidy
make proto-gen
make swagger
go work sync
npm update --save
```

Explaining in sequence:

1. Change to `proto` folder to update the lockfile. This will change the Buf lockfile references. **Check if any changes are necessary to the other Buf files**, i.e., `buf.yaml`, `buf.gen.gogo.yaml`, `buf.gen.pulsar.yaml` (impacts API folder usually), `buf.gen.swagger.yaml`. **Note**: if this changes, we should also execute/change on `ts-proto` once the release is done on develop/main here.
2. Run same sequence of commands that goreleaser would execute. In an ideal state is should come back clean with no changes, but in this case, there were changes, so those were checked in.
3. Since there's a change in `go.mod` / `go.sum` files, I ran `go work sync` as well.
4. Unrelated bumps, but also bumped NPM deps which are relevant for Semantic Release only.


ankurdotb:
Task linked: [DEV-5011 Protobuf lockfile update + regenerate](https://app.clickup.com/t/6600954/DEV-5011)

Eengineer1 (approved):
Approved 👍.

PR #901, opened 2025-06-11
perf: Delegate latest resource version check to client [DEV-5093]

DaevMithran:
This PR introduces an optional `previousVersionResourceId` field in `MsgCreateResourcePayload`, delegating the responsibility of fetching the latest resource version to the client when provided.

### ✅ Fixes

* Avoids out-of-gas errors on resource updates like status lists.
* Prevents incorrect fee estimation for a resource

###  Notes

* If `previousVersionResourceId` is omitted, the chain falls back to on-chain resolution.

This pull request introduces a new optional field, `previous_version_resource_id`, to the `MsgCreateResourcePayload` structure in the `cheqd.resource.v2` module. The changes ensure that this field is fully integrated into serialization, deserialization, and reflection logic, allowing users to specify a human-readable identifier for the resource's previous version. This addition enhances version tracking capabilities for resources.

### Addition of `previous_version_resource_id` field:

* **Field Definition**:
  - Added `previous_version_resource_id` as an optional string field to `MsgCreateResourcePayload`. This field allows specifying the identifier of the resource being updated, aiding in version tracking. [[1]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R2935-R2940) [[2]](diffhunk://#diff-03d103b382f11986ac5c6e65b3344938b192de47073a65a567927b0fea1d5f14R89-R98)
  - Updated the example payload in `tx_create_resource.go` to include `previousVersionResourceId` as an example field for CLI usage.

### Integration into Reflection Logic:

* **Reflection Field Descriptors**:
  - Added `fd_MsgCreateResourcePayload_previous_version_resource_id` to the list of field descriptors in `tx.pulsar.go`. This ensures the field is recognized during reflection-based operations. [[1]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R653) [[2]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R666)
  - Updated methods like `Range`, `Has`, `Clear`, `Get`, `Set`, and `Mutable` to handle the new field. These changes ensure proper serialization, deserialization, and field manipulation. [[1]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R776-R781) [[2]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R811-R812) [[3]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R843-R844) [[4]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R885-R887) [[5]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R924-R925) [[6]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R964-R965)

### Serialization and Deserialization Updates:

* **ProtoMethods**:
  - Modified `ProtoMethods` to include logic for serializing and deserializing `previous_version_resource_id`. This ensures the field is correctly encoded and decoded in protocol buffers. [[1]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R1095-R1098) [[2]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R1128-R1134) [[3]](diffhunk://#diff-950fe7cafe0c2eeace0903b3c457ced790c3c0226a27ed2d82509e58004c9951R1470-R1501)

### Updates to Protocol Buffers (`proto` file):

* **Protocol Buffers Definition**:
  - Updated `tx.proto` to define the `previous_version_resource_id` field with appropriate tags for JSON serialization and nullability.

These changes collectively enhance the `MsgCreateResourcePayload` structure, enabling better resource version management and integration with existing serialization and reflection mechanisms.


Eengineer1:
@DaevMithran Your recommended approach looks good. However, we could go with a solution that removes tracking and persisting resource version identifiers on the application / client-side (both of which this approach assumes):

### Maintain an Index for Latest Version

Whenever a new version is added, update a dedicated key (e.g., latest:<collectionID>:<name>:<resourceType>) that always points to the latest version’s metadata or ID. Then, to get the latest version, simply fetch this record—no walking required.

Implementation outline:

    On AddNewResourceVersion, always update the “latest” index.
    When querying, always read the “latest” index first.

Example:

```go
func (k Keeper) AddNewResourceVersion(ctx context.Context, resource *types.ResourceWithMetadata) error {
    // ... existing logic ...
    // After adding the resource:
    latestKey := collections.Join(resource.Metadata.CollectionId, resource.Metadata.Name, resource.Metadata.ResourceType, "latest")
    err := k.LatestResourceIndex.Set(ctx, latestKey, resource.Metadata)
    if err != nil {
        return err
    }
    // ... continue as before ...
}

func (k Keeper) GetLastResourceVersionMetadata(ctx context.Context, collectionID, name, resourceType string) (types.Metadata, bool, error) {
    latestKey := collections.Join(collectionID, name, resourceType, "latest")
    latest, err := k.LatestResourceIndex.Get(ctx, latestKey)
    if err != nil {
        return types.Metadata{}, false, err
    }
    return latest, true, nil
}
```

Now, no walking and no client-supplied previous version are needed. Since we already maintain PreviousVersionId and NextVersionId in your metadata, we only need a way to quickly get to the head/tail of the chain (the latest version). That’s what the index above solves.

DaevMithran:
> @DaevMithran Your recommended approach looks good. However, we could go with a solution that removes tracking and persisting resource version identifiers on the application / client-side (both of which this approach assumes):
> 
> ### Maintain an Index for Latest Version

@Eengineer1 I mentioned this approach in the doc https://app.clickup.com/6600954/docs/69e7u-14952/69e7u-28072. But we have to perform migrations for all the existing resources right (Isn't that expensive?) ? If we do not perform a migration the resources which are throwing out of gas errors on resource updates will continue to do so.

Eengineer1:
> > @DaevMithran Your recommended approach looks good. However, we could go with a solution that removes tracking and persisting resource version identifiers on the application / client-side (both of which this approach assumes):
> > ### Maintain an Index for Latest Version
> 
> @Eengineer1 I mentioned this approach in the doc https://app.clickup.com/6600954/docs/69e7u-14952/69e7u-28072. But we have to perform migrations for all the existing resources right (Isn't that expensive?) ? If we do not perform a migration the resources which are throwing out of gas errors on resource updates will continue to do so.

Correct, we will add the migrations necessary. That's mandatory.

Regarding computational expenses, we're already migrating store keys in this next release iteration, so this may be an added step as well as expediting this performance fix to include it in this release.

PR #902, opened 2025-06-15
chore: fix function name

cuithon:
fix function name


tweeddalex:
@Eengineer1 this is a commit from an external dev. Can we resolve this one?

